PR #15, opened 2021-07-26
Feature/ar 575 create slide 07 2021

sinejespersen:
https://jira.itkdev.dk/browse/AR-575

* added images (that are opensource)
* added and edited fixture data
* added edit slide
* renamed some variables from snakecase to camelcase
* changed how validation works
* created a component that renders form elements from json
* added form checkbox
* added dropdown to navbar
* added image uploader
* added select playlist table for selecting a playlist and viewing the selected in a table

image

image

image


cableman (approved):
The fixture images, do we have the right permissions to use them and upload theme here?

PR #18, opened 2021-08-01
Feature/ar 576 upload image 07 2021

sinejespersen:
https://jira.itkdev.dk/browse/AR-576

image

image

image

* changed from react intl to i18next
* added edit media
* changed the image uploader to accept name/description at all time (I am assuming this makes sense, as all images need and alt text and a name?)


turegjorup (approved):
Approved. But this "changed from react intl to i18next" should have been done in an individual PR

PR #26, opened 2021-09-29
Media upload

cableman:
Add media upload end-points

Screenshot 2021-09-30 at 15 22 53


src/Controller/MediaController.php
```diff
@@ -0,0 +1,33 @@
...
+
+        $media = new Media();
+        $media->setFile($uploadedFile);
+        $media->setTitle($request->request->get('title'))
+            ->setDescription($request->request->get('description'))
+            ->setLicense($request->request->get('license'))
+            ->setModifiedBy($request->request->get('modifiedBy'))
+            ->setCreatedBy($request->request->get('createdBy'));
+
+        // Note that the extra information about the uploaded file is added in the MediaPrePersistEventListener as first
```
tuj, line 29: "as first there the file is on disk."
changed to
"because then the file exists on disk."

tuj (approved):

PR #45, opened 2021-11-02
AR-707: Added command to load templates

tuj:
#### Link to ticket

https://jira.itkdev.dk/browse/AR-717

#### Description

Adds command to import a template from a .json file.

#### Checklist


turegjorup (approved):
This works for now.

Later we should extract the json schema to it's own file. The entire point in using json schema is that having a schema enables us to validate templates various places. I.E we could set up github actions to validate against the schema in PR's. Or we could make a template repository for making new Templates, with tooling build in that validates against the schema to help with DX. Having the schema embedded in the PHP command prevents this.

PR #46, opened 2021-11-03
Add microseconds to datetime format to support format given by EcmaSc…

turegjorup:
Add microseconds to datetime format to support format given by EcmaScript Date.prototype.toISOString()

#### Link to ticket

https://jira.itkdev.dk/browse/AR-725

#### Description

EcmaScript has `Date.prototype.toISOString()`([link](https://developer.mozilla.org/en-US/docs/Web/JavaScript/Reference/Global_Objects/Date/toISOString)) which gives a string containing microseconds: `2011-10-05T14:48:00.000Z`.

To support this we change the default api datetime format to match the js iso string. This is also supported by OpenAPI whih follows [RFC3339](https://datatracker.ietf.org/doc/html/rfc3339#section-5.6): 
```
partial-time    = time-hour ":" time-minute ":" time-second
                     [time-secfrac]
```

#### Screenshot of the result

Screenshot 2021-11-03 at 13 04 40

#### Checklist

### DIFF
--- a/docker-compose.server.yml
+++ b/docker-compose.server.yml
@@ -22,7 +22,7 @@ services:
       - APP_SECRET=MySuperSecret
       - APP_CORS_ALLOW_ORIGIN='^https?://localhost(:[0-9]+)?$'
       - APP_DATABASE_URL=mysql://db:db@mariadb:3306/db?serverVersion=mariadb-10.4.0
-      - APP_DEFAULT_DATE_FORMAT='Y-m-d\TH:i:s:v\Z'
+      - APP_DEFAULT_DATE_FORMAT='Y-m-d\TH:i:s.v\Z'
 
   nginx:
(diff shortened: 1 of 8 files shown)


docker-compose.server.yml
```diff
@@ -22,7 +22,7 @@ services:
       - APP_SECRET=MySuperSecret
       - APP_CORS_ALLOW_ORIGIN='^https?://localhost(:[0-9]+)?$'
       - APP_DATABASE_URL=mysql://db:db@mariadb:3306/db?serverVersion=mariadb-10.4.0
-      - APP_DEFAULT_DATE_FORMAT='Y-m-d\TH:i:s\Z'
+      - APP_DEFAULT_DATE_FORMAT='Y-m-d\TH:i:s:v\Z'
```
cableman, line 25: `:v` should be `.v`

cableman (approved):

PR #53, opened 2021-12-06
Feature/ar 754 commands updated 12 2021

sinejespersen:
https://jira.itkdev.dk/browse/AR-754

### DIFF
--- a/src/Command/LoadScreenLayoutsCommand.php
+++ b/src/Command/LoadScreenLayoutsCommand.php
@@ -11,6 +11,7 @@ use Symfony\Component\Console\Input\InputArgument;
 use Symfony\Component\Console\Input\InputInterface;
 use Symfony\Component\Console\Output\OutputInterface;
 use Symfony\Component\Console\Style\SymfonyStyle;
+use Symfony\Component\Uid\Ulid;
 
 #[AsCommand(
     name: 'app:screen-layouts:load',
@@ -36,8 +37,17 @@ class LoadScreenLayoutsCommand extends Command
         if ($filename = $input->getArgument('filename')) {
             try {
                 $content = json_decode(file_get_contents($filename), false, 512, JSON_THROW_ON_ERROR);
+                $loadedScreenLayout = $this->entityManager->getRepository(ScreenLayout::class)->findBy(['id' => Ulid::fromString($content->id)]);
 
-                $screenLayout = new ScreenLayout();
+                if (is_array($loadedScreenLayout) & 0 === count($loadedScreenLayout)) {
+                    // If the screen layout doesnt exist, a new will be created
+                    $screenLayout = new ScreenLayout();
+                    $screenLayout->setId(Ulid::fromString($content->id));
+                } else {
+                    // If the screen layout already exists it will be replaced
+                    $screenLayout = array_shift($loadedScreenLayout);
+                }
+
                 $screenLayout->setTitle($content->title);
                 $screenLayout->setGridColumns($content->grid->columns);
                 $screenLayout->setGridRows($content->grid->rows);
--- a/src/Command/LoadScreenLayoutsCommand.php
+++ b/src/Command/LoadScreenLayoutsCommand.php
@@ -5,6 +5,7 @@ namespace App\Command;
 use App\Entity\ScreenLayout;
 use App\Entity\ScreenLayoutRegions;
 use Doctrine\ORM\EntityManagerInterface;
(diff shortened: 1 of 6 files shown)


src/Command/LoadScreenLayoutsCommand.php
```diff
@@ -37,7 +38,23 @@ final protected function execute(InputInterface $input, OutputInterface $output)
             try {
                 $content = json_decode(file_get_contents($filename), false, 512, JSON_THROW_ON_ERROR);
 
-                $screenLayout = new ScreenLayout();
+                if (isset($content->id)) {
```
turegjorup, line 41: You should either add `&& Ulid::isValid($content->id)` to the `if`clause or catch the `\InvalidArgumentException` thrown by `Ulid::fromString`on inavlid value

src/Command/LoadScreenLayoutsCommand.php
```diff
@@ -37,7 +38,23 @@ final protected function execute(InputInterface $input, OutputInterface $output)
...
+                    $loadedScreenLayout = $this->entityManager->getRepository(ScreenLayout::class)->findBy(['id' => Ulid::fromString($content->id)]);
```
turegjorup, line 42: Using `findOneBy` will give you a `ScreenLayout` instance or `null`. Then you can do a simple null check instead of the `is_array`, `count`,  `array_shift`

src/Command/LoadTemplateCommand.php
```diff
@@ -59,7 +60,22 @@ final protected function execute(InputInterface $input, OutputInterface $output)
                     return Command::INVALID;
                 }
 
-                $template = new Template();
+                if (isset($content->id)) {
+                    $loadedTemplate = $this->entityManager->getRepository(Template::class)->findBy(['id' => Ulid::fromString($content->id)]);
```
turegjorup, line 64: Same as above

turegjorup (changes requested):
The changes to the entities are not needed. They both `use EntityIdTrait;` where `Ulid` and getter/setter is defined.

Also if you ever need to actually add fields to entities you must make a doctrine migration to update the database schema:
`bin/console doctrine:migrations:diff` then sanity check the generated file

src/Command/LoadScreenLayoutsCommand.php
```diff
@@ -26,43 +28,59 @@ public function __construct(
 
     protected function configure(): void
     {
-        $this->addArgument('filename', InputArgument::REQUIRED, 'json file to load');
+        $this->addArgument('filename', InputArgument::REQUIRED, 'json file to load. Can be local file or a URL');
```
tuj, line 31: "Can be a local file or an URL"

src/Command/LoadScreenLayoutsCommand.php
```diff
@@ -26,43 +28,59 @@ public function __construct(
...
-                foreach ($content->regions as $localRegion) {
-                    $region = new ScreenLayoutRegions();
-                    $region->setGridArea($localRegion->gridArea);
-                    $region->setTitle($localRegion->title);
-                    $this->entityManager->persist($region);
-                    $screenLayout->addRegion($region);
-                }
+        try {
+            $filename = $input->getArgument('filename');
+            $content = json_decode(file_get_contents($filename), false, 512, JSON_THROW_ON_ERROR);
```
tuj, line 40: Adding a json schema validation to make sure title, grid, etc. are correctly formatted would ensure data integrity.

See:
https://github.com/os2display/display-api-service/blob/develop/src/Command/LoadTemplateCommand.php#L43

src/Command/LoadTemplateCommand.php
```diff
@@ -28,58 +30,72 @@ public function __construct(
 
     protected function configure(): void
     {
-        $this->addArgument('filename', InputArgument::REQUIRED, 'json file to load');
+        $this->addArgument('filename', InputArgument::REQUIRED, 'json file to load. Can be local file or a URL');
```
tuj, line 33: "Can be a local file or an URL"

tuj (changes requested):

src/Command/LoadScreenLayoutsCommand.php
turegjorup: Agreed. And we have a TODO as to where these schemas are stored and exposed. I like the linked example, but inclining the schema in the command file prevents reuse. 

Afraid it's out of scope right now

turegjorup (approved):

tuj (approved):

PR #59, opened 2022-01-05
Feature/ar 551 koba

tuj:
#### Link to ticket

https://jira.itkdev.dk/browse/AR-770

#### Description

Adds support for koba resources.

#### Screenshot of the result

Screenshot 2022-01-05 at 13 34 29

#### Checklist


src/Feed/KobaFeedType.php
```diff
@@ -0,0 +1,135 @@
+<?php
+
+namespace App\Feed;
```
turegjorup, line 3: Have we given any thought on how to namespace/install custom feeds? KOBA is specific to Aarhus....

turegjorup (approved):

src/Feed/KobaFeedType.php
tuj: I think this could be solved by just adding a bundle where the feed types implement FeedTypeInterface.
https://github.com/os2display/display-api-service/blob/develop/config/services.yaml#L19
The issue is how to be able to access this interface from the bundle.

PR #65, opened 2022-01-27
Feature/ar 721 campaign 2022 01

sinejespersen:
https://jira.itkdev.dk/browse/AR-721


src/Dto/Playlist.php
```diff
@@ -12,8 +15,17 @@ class Playlist
     public string $modifiedBy = '';
     public string $createdBy = '';
     public string $slides = '';
+    public Collection $campaignScreens;
```
tuj, line 18: Should be array in DTO

tuj (changes requested):

tuj (approved):

PR #66, opened 2022-01-27
Feature/ar 647

tuj:
#### Link to ticket

https://jira.itkdev.dk/browse/AR-647

#### Description

Adds endpoints for binding machines to screen entities and for authenticating a screen.

#### Screenshot of the result

Screenshot 2022-01-27 at 11 20 52

Screenshot 2022-01-27 at 11 20 11

Screenshot 2022-01-28 at 08 56 17

#### Checklist


config/packages/cache.yaml
```diff
@@ -16,3 +16,20 @@ framework:
...
                 # Default expire set to 5 minutes
                 default_lifetime: 300
+
+            # Creates a "auth-screen.cache" service
+            authscreen.cache:
+                adapter: cache.adapter.redis
+                # Default expire set to 1 hour
+                default_lifetime: 3600
+
+when@test:
```
turegjorup, line 26: We should decide when/if to use this as opposed to having independent yaml files i.e `/config/packages/test/cache.yaml`

config/packages/cache.yaml
```diff
@@ -16,3 +16,20 @@ framework:
                 adapter: cache.adapter.redis
                 # Default expire set to 5 minutes
                 default_lifetime: 300
+
+            # Creates a "auth-screen.cache" service
+            authscreen.cache:
```
turegjorup, line 21: This can be autowired as `$authscreenCache` (https://symfony.com/doc/current/cache.html#creating-custom-namespaced-pools)

config/services.yaml
```diff
@@ -34,6 +37,10 @@ services:
             - !tagged_iterator app.feed.feed_type
             - '@feeds.cache'
 
+    App\Service\AuthScreenService:
+        arguments:
+            - '@authscreen.cache'
```
turegjorup, line 42: Not needed if you follow the rules for autowireable naming (https://symfony.com/doc/current/cache.html#creating-custom-namespaced-pools)

public/api-spec-v1.json
```diff
@@ -0,0 +1,6554 @@
+{
```
turegjorup: Not sure we need/want the json spec as well. The composer command and github actions only generate and check the yaml api spec. So theis file can become outdated without us knowing

src/Service/AuthScreenService.php
```diff
@@ -0,0 +1,136 @@
...
+use Lexik\Bundle\JWTAuthenticationBundle\Services\JWTTokenManagerInterface;
+use Symfony\Component\Uid\Ulid;
+use Symfony\Contracts\Cache\CacheInterface;
+
+class AuthScreenService
+{
+    public const BIND_KEY_PREFIX = 'BindKey-';
+
+    public function __construct(
+        private CacheInterface $cache,
```
turegjorup, line 17: Fix parameter naming for autowiring (https://symfony.com/doc/current/cache.html#creating-custom-namespaced-pools)

src/Service/AuthScreenService.php
```diff
@@ -0,0 +1,136 @@
...
+
+                    // Remove bindKey entry.
+                    $this->cache->delete(AuthScreenService::BIND_KEY_PREFIX.$bindKey);
+
+                    return true;
+                }
+            }
+        }
+
+        return false;
```
turegjorup, line 112: I'm not a fan of returning `false` when something fails. I know this is a standard PHP pattern, but I really think this should be an exception, This is not something that should fail under normal conditions.

Also using the `true/false return` pattern can lead to that is hard to consume, see "No more success result booleans" on https://flysystem.thephpleague.com/v2/docs/what-is-new/

src/Service/AuthScreenService.php
```diff
@@ -0,0 +1,136 @@
...
+    }
+
+    public function unbindScreen(Screen $screen): void
```
turegjorup, line 115: This is inconsistent with `bindScreen()`. Why return `bool` from `bindScreen()` and `void` from `unbindScreen()`

src/Controller/AuthScreenBindController.php
```diff
@@ -0,0 +1,41 @@
...
+        $screen = $this->screenRepository->find($screenUlid);
+
+        $body = $request->toArray();
+        $bindKey = $body['bindKey'];
+
+        if (!isset($bindKey)) {
+            throw new \HttpException('Missing key', 400);
+        }
+
+        $success = $this->authScreenService->bindScreen($screen, $bindKey);
```
turegjorup, line 33: This should be wrapped in `try/catch`, see comment for the `bindScreen()` function

turegjorup (changes requested):

config/packages/cache.yaml
tuj: fixed

src/Service/AuthScreenService.php
tuj: Fixed

src/Service/AuthScreenService.php
tuj: Will change bindScreen

src/Controller/AuthScreenBindController.php
tuj: Fixed

public/api-spec-v1.json
tuj: We use the json spec for generating apiSlices in the admin. I will add it to github actions validation steps

tuj: Have already added it to composer command: https://github.com/os2display/display-api-service/pull/66/files#diff-d2ab9925cad7eac58e0ff4cc0d251a937ecf49e4b6bf57f8b95aab76648a9d34R132

turegjorup (approved):

PR #74, opened 2022-02-09
AR-781: add playlist to slide

sinejespersen:
#### Link to ticket

https://jira.itkdev.dk/browse/AR-781

#### Description

Please include a short description of the suggested change and the reasoning behind the approach you have chosen.

#### Screenshot of the result

#### Checklist

### DIFF
--- a/src/Repository/PlaylistSlideRepository.php
+++ b/src/Repository/PlaylistSlideRepository.php
@@ -37,11 +37,11 @@ class PlaylistSlideRepository extends ServiceEntityRepository
         $firstResult = ($page - 1) * $itemsPerPage;
 
         $queryBuilder = $this->_em->createQueryBuilder();
+        $queryBuilder2 = $this->_em->createQueryBuilder();
         $queryBuilder->select('s')
-            ->from(Slide::class, 's')
-            ->innerJoin('s.playlistSlides', 'ps', Join::WITH, 'ps.playlist = :playlistId')
-            ->setParameter('playlistId', $playlistUid, 'ulid')
-            ->orderBy('ps.weight', 'ASC');
+            ->from(Playlist::class, 's')
+            ->innerJoin('s.playlistSlides', 'ps', Join::WITH, 'ps.slide = :slideId')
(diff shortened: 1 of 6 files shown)


src/Repository/PlaylistSlideRepository.php
```diff
@@ -114,6 +113,57 @@ public function updateRelations(Ulid $playlistUlid, ArrayCollection $collection)
...
+                $playlist = $playlistRepos->findOneBy(['id' => $entity->playlist]);
+                if (is_null($playlist)) {
+                    throw new InvalidArgumentException('Playlist not found');
+                }
+
+                // Create new relation.
+                $ps = new PlaylistSlide();
+                $ps->setPlaylist($playlist)
+                    ->setSlide($slide)
+                    ->setWeight(0);
```
tuj, line 152: Shouldn't the weight change according to where the slide should occur in the playlist? 
So you cannot just remove all the existing relations and just make them anew with weight 0. This will change playlist order.

tuj (changes requested):

src/Repository/PlaylistSlideRepository.php
```diff
@@ -114,6 +114,73 @@ public function updateRelations(Ulid $playlistUlid, ArrayCollection $collection)
...
+        }
+
+        $playlistRepos = $this->entityManager->getRepository(Playlist::class);
+
+        $this->entityManager->getConnection()->beginTransaction();
+        try {
+            // Remove all existing relations between slides and current playlist.
+            $entities = $this->findBy(['slide' => $slide]);
+            foreach ($entities as $entity) {
+                $this->entityManager->remove($entity);
```
tuj, line 132: This will change the order of slides in alle the playlists the slide continues to be present. They should only be remove if the do not occur in playlist anymore.

tuj (changes requested):
Please add a test case that confirms the expected behaviour for updateSlidePlaylistRelations(Ulid $slideUlid, ArrayCollection $collection).

It should confirm that if a slide is added to several playlists, and then removed from one playlist (in collection) the slide should have the same weight in the playlists it has not been removed from.

PR #79, opened 2022-03-01
Feature/ar 700 tenants

turegjorup:
#### Link to ticket

https://jira.itkdev.dk/browse/AR-700

#### Description

Implemented the Tenant concept. Notes:

* All Migrations are removed and replaced with one new migration
*  All Tenant scope entities are now under the `App\Entity\Tenant` namespace. This change is the cause of the large number of change files.
* Removed `stof/doctrine-extensions-bundle`. Using simple lifecycle events instead.
* Refactored the entity model from traits to one abstract base class to better support life cycle methods
* `app:make:user` command updated to account for tenants but just gives access to all tenants right now.
* Tests have been minimally adapted so they pass.

ToDo:

- [x] Update OpenApi spec with tenant concept
- [ ] Update test suite to properly test that content is created with active tenant, and that users can't see content from tenants they don't have access to.
- [ ] Update Readme with proper description of how the Tenant concept is implemented
- [ ] Add a Doctrine filter to ensure that content is filtered to active tenant

#### Screenshot of the result

No screenshot

#### Checklist


src/Command/AddUserCommand.php
```diff
@@ -167,12 +170,19 @@ protected function execute(InputInterface $input, OutputInterface $output): int
         $user = new User();
         $user->setEmail($email);
         $user->setFullName($fullName);
-        $user->setRoles([$isAdmin ? 'ROLE_ADMIN' : 'ROLE_USER']);
 
         // See https://symfony.com/doc/5.4/security.html#registering-the-user-hashing-passwords
         $hashedPassword = $this->passwordHasher->hashPassword($user, $plainPassword);
         $user->setPassword($hashedPassword);
 
+        $tenants = $this->tenants->findAll();
```
tuj, line 178: Please add a TODO message that tenants should be set by the user.

src/Service/TenantFactory.php
```diff
@@ -0,0 +1,44 @@
...
+use Doctrine\ORM\Query\QueryException;
+
+class TenantFactory
+{
+    public function __construct(private EntityManagerInterface $entityManager, private TenantRepository $tenantRepository)
+    {
+    }
+
+    /**
+     * Get Tenants from array og tenent keys. Create new Tenants
```
tuj, line 17: og = of

src/Service/TenantFactory.php
```diff
@@ -0,0 +1,44 @@
...
+     * for unknown keys.
+     *
+     * @param array $tenantKeys
+     *
+     * @return array
+     *
+     * @throws QueryException
+     */
+    public function getTenants(array $tenantKeys): array
```
tuj, line 26: getTenants is a bit weird naming since it has side-effects. Maybe setupTenants...

tuj (approved):

tuj (approved):

PR #101, opened 2022-03-25
AR-611: screens from screen group

sinejespersen:


src/Repository/ScreenRepository.php
```diff
@@ -17,5 +21,23 @@ class ScreenRepository extends ServiceEntityRepository
     public function __construct(ManagerRegistry $registry)
     {
         parent::__construct($registry, Screen::class);
+        $this->entityManager = $this->getEntityManager();
```
tuj, line 24: Why add this?

tuj (approved):

PR #103, opened 2022-03-31
AR-544: remove filter add new filter

sinejespersen:
https://jira.itkdev.dk/browse/AR-544

* added filter "shared with me"
* removed doctrine filter
* rewrote controllers to dataproviders and added them to services.yml


tests/Api/PlaylistsTest.php
```diff
@@ -365,9 +362,25 @@ public function testGetScreensList(): void
...
         $this->assertJsonContains([
-            '@context' => '/contexts/Screen',
-            '@id' => '/v1/screens',
+            '@context' => '/contexts/ScreenCampaign',
+            '@id' => '/v1/screen-campaigns',
+            '@type' => 'hydra:Collection',
+        ]);
+    }
+
+    public function testSharedPlaylists(): void
```
tuj, line 371: I think what we are missing in terms of tests, is that we can CRUD through the API. And that things are as expected after the different operations.

src/DataProvider/MediaItemDataProvider.php
```diff
@@ -0,0 +1,69 @@
...
+
+        // If there is not a result, shared playlists should be checked.
+        if (is_null($media)) {
+            $connectedSlides = $this->slideRepository->getSlidesByMedia($id)->getQuery()->getResult();
+            foreach ($connectedSlides as $slide) {
+                $playlists = $this->playlistSlideRepository->getPlaylistsFromSlideId($slide->getId())->getQuery()->getResult();
+                foreach ($playlists as $playlist) {
+                    if (in_array($tenant, $playlist->getPlaylist()->getTenants()->toArray())) {
+                        $media = $this->mediaRepository->find($mediaUlid);
+                        break;
```
tuj, line 61: This should be `break 2;` to exist the double loop.

tuj (approved):

PR #126, opened 2023-01-02
AR-890: Added config option for setting token TTL for screen users

turegjorup:
#### Link to ticket

https://jira.itkdev.dk/browse/AR-890

#### Description

Added config option for setting token TTL for screen users

#### Screenshot of the result

#### Checklist

### DIFF
--- a/src/Security/ScreenAuthenticator.php
+++ b/src/Security/ScreenAuthenticator.php
@@ -17,8 +17,8 @@ class ScreenAuthenticator
     public const BIND_KEY_PREFIX = 'BindKey-';
 
     public function __construct(
-        private int $jwtRefreshTokenTtl,
-        private CacheInterface $authscreenCache,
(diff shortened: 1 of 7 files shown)


src/Security/ScreenAuthenticator.php
```diff
@@ -153,7 +153,7 @@ public function unbindScreen(Screen $screen): void
     private function generateBindKey(): string
     {
         $length = 8;
-        $chars = '123456789ABCDEFGHIJKLMNPQRSTUVWXYZ';
```
tuj, line 156: Why this change?

tuj (approved):

src/Security/ScreenAuthenticator.php
turegjorup: Having just numbers vs. numbers and chars makes no real difference from a security perspective. But having (capital) letters in the code is an inconvenience when entering the code.

PR #150, opened 2023-06-16
 DISPLAY-986: Update docker build and github workflows

turegjorup:
#### Link to ticket

https://jira.itkdev.dk/browse/DISPLAY-986

#### Description

* Update docker build to publish to "os2display" org on docker hup.
* Update docker build to use generic php image
* Update github workflow to latest actions.
* Add github workflow to build and create release.

#### Screenshot of the result

#### Checklist


infrastructure/display-api-service/docker-entrypoint.sh
```diff
@@ -2,6 +2,14 @@
 
 set -eux
 
+## Start the PHP FPM process.
+echo "Starting PHP 8.1 FPM"
+
+# first arg is `-f` or `--some-option`
+if [ "${1#-}" != "$1" ]; then
+  set -- php-fpm "$@"
```
cableman, line 10: If you run PHP here it will block the reset of the execution of this script. Should be moved to just to the linjes before the last line.

infrastructure/display-api-service/gitconfig
```diff
@@ -0,0 +1,9 @@
+[color]
```
cableman, line 1: Actually only the "[safe]" config is required to make things work.

cableman (changes requested):

infrastructure/display-api-service/docker-entrypoint.sh
turegjorup: Moved

infrastructure/display-api-service/gitconfig
turegjorup: Will leave in just the same. Can see we do stuff like `%C(yellow)%d%Creset %s %C(bold blue)<%an>%Creset`

PR #152, opened 2023-06-29
Bugfix/nginx entrypoint failure fixed

inuitviking:
#### Description

NGINX had issues with the entrypoint. This fixes that.
As well as redirecting to a new file does not always work, so redirecting to stderr instead; perhaps this isn't the best option, but fixes the issue at least temporarily.

#### Checklist

If anything needs to be changed, please do mention it, and I will correct it if I can.


cableman:
We have fixed this in https://github.com/os2display/display-api-service/pull/153/files

PR #154, opened 2023-07-07
DISPLAY-986: Updated add user command to ask which tenants user belongs to

turegjorup:
#### Link to ticket

https://jira.itkdev.dk/browse/SUPP0RT-1109

#### Description

Updated add user command to ask which tenants user belongs to

#### Screenshot of the result

#### Checklist


CHANGELOG.md
```diff
@@ -3,13 +3,14 @@
 All notable changes to this project will be documented in this file.
 
 ## [Unreleased]
-
+- Update docker build to publish to "os2display" org on docker hup. Update github workflow to latest actions.
```
rimi-itk, line 6: Typo: “hup” should be “hub”

src/Command/User/AddUserCommand.php
```diff
@@ -80,7 +81,8 @@ protected function configure(): void
             ->addArgument('email', InputArgument::OPTIONAL, 'The email of the new user')
             ->addArgument('password', InputArgument::OPTIONAL, 'The plain password of the new user')
             ->addArgument('full-name', InputArgument::OPTIONAL, 'The full name of the new user')
-            ->addOption('admin', null, InputOption::VALUE_NONE, 'If set, the user is created as an administrator')
+            ->addArgument('role', InputArgument::OPTIONAL, 'The role of the user [editor|admin]')
+            ->addArgument('tenant-keys', InputArgument::IS_ARRAY | InputArgument::OPTIONAL, 'The keys of the tenants the user should belong to (separate multiple keys with a space)')
```
rimi-itk, line 85: I would use options for this (rather than arguments) to make it easier to handle multiple values, e.g.

```sh
app:user:add … --role=admin --tenant=hat --tenant=briller
```

but that's a matter of opinion.

src/Command/User/AddUserCommand.php
```diff
@@ -149,6 +156,39 @@ protected function interact(InputInterface $input, OutputInterface $output): voi
...
+
+        $helper = $this->getHelper('question');
+
+        // Ask for the role if it's not defined
+        $role = $input->getArgument('role');
+        if (null !== $role) {
+            $this->io->text(' > <info>Role</info>: '.$role);
+        } else {
+            $question = new ChoiceQuestion(
+                'Please select the users role (defaults to editor)',
```
rimi-itk, line 168: Typo: “users” should be “user's”.

src/Command/User/AddUserCommand.php
```diff
@@ -149,6 +156,39 @@ protected function interact(InputInterface $input, OutputInterface $output): voi
...
+                ['editor', 'admin'],
+                0
+            );
+            $question->setErrorMessage('Color %s is invalid.');
```
rimi-itk, line 172: Yes, but what about the role?!

src/Command/User/AddUserCommand.php
```diff
@@ -149,6 +156,39 @@ protected function interact(InputInterface $input, OutputInterface $output): voi
...
+            $output->writeln('You have just selected: '.$role);
+        }
+
+        // Ask for the tenant keys if it's not defined
+        $tenantKeys = $input->getArgument('tenant-keys');
+        if (0 < \count($tenantKeys)) {
+            $this->io->text(' > <info>Tenant Keys</info>: '.$tenantKeys);
+        } else {
+            $question = new ChoiceQuestion(
+                'Please select the tenant(s) the user should belong to (to select multiple answer with a list. E.g: "key1, key3")',
```
rimi-itk, line 184: I don't understand what “(to select multiple answer with a list. E.g: "key1, key3")” means. Doesn't the helper help the user picking multiple values?

src/Command/User/AddUserCommand.php
```diff
@@ -226,17 +279,12 @@ private function getCommandHelp(): string
...
-
-  <info>php %command.full_name%</info> email password <comment>--admin</comment>
+  <info>php %command.full_name%</info> <comment>email password fullname role tenants</comment>
 
-If you omit any of the two required arguments, the command will ask you to
+If you omit any of the required arguments, the command will ask you to
 provide the missing values:
 
-  # command will ask you for the password
+  # command will ask you for the password etc
```
rimi-itk, line 287: Add period after “etc”.

src/Utils/CommandInputValidator.php
```diff
@@ -83,4 +88,39 @@ public function validateFullName(?string $fullName): string
...
         return $fullName;
     }
+
+    public function validateRole(?string $role): string
+    {
+        if (empty($role)) {
+            throw new InvalidArgumentException('The role can not be empty.');
+        }
+
+        $allowedRoles = ['editor', 'admin'];
```
rimi-itk, line 98: For future-proofing you could set these roles in an environment variable. Or at least put the list into a class variable; it's also used in `interact`.

rimi-itk (changes requested):

src/Command/User/AddUserCommand.php
turegjorup: No. You have to answer with a list of comma separated values. There is no "UI".

src/Utils/CommandInputValidator.php
turegjorup: Moved to a class constant.

Originally I tried to get the values through DI and the one of the Security services. Didn't work as I wanted because it's a subset of the defined roles we allow here.

I don't like moving them to `env` because they would still have to match what is defined in `security.yaml`. Also I consider `env` values to be something you can configure. But Symfony's docs encourage you to do stuff like `#[IsGranted('ROLE_ADMIN')]` in code, so having the roles configurable would clash with that.

src/Command/User/AddUserCommand.php
turegjorup: https://symfony.com/doc/current/components/console/helpers/questionhelper.html#multiple-choices

```
use Symfony\Component\Console\Question\ChoiceQuestion;

// ...
public function execute(InputInterface $input, OutputInterface $output): int
{
    // ...
    $helper = $this->getHelper('question');
    $question = new ChoiceQuestion(
        'Please select your favorite colors (defaults to red and blue)',
        ['red', 'blue', 'yellow'],
        '0,1'
    );
    $question->setMultiselect(true);

    $colors = $helper->ask($input, $output, $question);
    $output->writeln('You have just selected: ' . implode(', ', $colors));

    return Command::SUCCESS;
}
```

turegjorup: Notice `implode(', ', $colors)`

turegjorup: But yes, before testing it I also expected a checkbox like feature

src/Command/User/AddUserCommand.php
rimi-itk: And that example uses an option, `--admin`, for (implicitly) setting the role. This is not really important, so let's leave it as it is for now.

rimi-itk (approved):

src/Command/User/AddUserCommand.php
turegjorup: Agreed 😄 

Changed because 
1) there might be more than 2 roles at some point, 
2) none of the people who has had to use the command have figured out the `--admin` option. So tried to make it more obvious how to select role

PR #162, opened 2023-09-01
Support external users

tuj:
#### Link to ticket

https://jira.itkdev.dk/browse/DISPLAY-993

#### Description

* Adds "external" openid-connect provider.
* Renamed "oidc" openid-connect provider to "ad".
* Modifies User to support external user type.
* Adds command to set user type.
* Expands api with external user endpoints.
* Upgrades openid-connect bundle to 3.1 to support multiple providers.
* Changes php requirement in composer.json to >= 8.1.
* Changed user identifier to providerId.

NB! Uses the changes from https://github.com/os2display/display-api-service/pull/163 since support for multiple providers is necessary.

#### Screenshots

Screenshot 2023-10-04 at 12 17 01

#### Checklist


turegjorup (changes requested):
General comments:

* Command to change user type can be removed. When changing DB schema I consider it a to be a valid option to set values for new columns in the migration adding the coloumn
* https://github.com/os2display/display-api-service/pull/163 should be finished and merged. Then this PR should be updated
* Email field should not be nullable
* We should introduce a `$providerID` on the `User` entity given that we don't get an email from MitID
* Consider setting users fullname from MitID or asking user when activating

src/Controller/UserActivateController.php
```diff
@@ -0,0 +1,33 @@
...
+
+    public function __invoke(Request $request): JsonResponse
+    {
+        $body = json_decode($request->getContent());
+        $activationCode = $body->activationCode;
+
+        try {
+            $this->externalUserService->activateExternalUser($activationCode);
+        } catch (ExternalUserCodeException $e) {
+            throw new HttpException($e->getCode(), $e->getMessage());
```
turegjorup, line 28: On the fence on the exception here. The code given is important for the API client to understand the error. Is the the controller or the `ExternalUserService` who should choose the proper code? Doing it this way pushes the responsibility for choosing the correct HTTP code down to the service. But I think it's a bit of a code smell to have the `ExternalUserService`be required to know HTTP semantics.

src/Controller/UserRefreshCodeController.php
```diff
@@ -0,0 +1,38 @@
...
+class UserRefreshCodeController extends AbstractController
+{
+    public function __construct(
+        private readonly UserActivationCodeRepository $activationCodeRepository,
+        private readonly ExternalUserService $externalUserService,
+        private readonly ValidationUtils $validationUtils,
+    ) {}
+
+    /**
+     * @throws CodeGenerationException
```
turegjorup, line 23: This should be caught and converted to the relevant `HttpException`

src/Controller/UserRefreshCodeController.php
```diff
@@ -0,0 +1,38 @@
...
+     * @throws \Exception
+     */
+    public function __invoke(Request $request, string $id)
+    {
+        $ulid = $this->validationUtils->validateUlid($id);
+
+        $code = $this->activationCodeRepository->find($ulid);
+
+        if (null === $code) {
+            throw new \Exception('Not found', 404);
```
turegjorup: This should be a `HttpException`

src/DataTransformer/ExternalUserActivationCodeInputDataTransformer.php
```diff
@@ -0,0 +1,78 @@
...
+            $roles[] = 'ROLE_EXTERNAL_USER_ADMIN';
+        } else {
+            $roles[] = 'ROLE_EXTERNAL_USER';
+        }
+
+        $code = new UserActivationCode();
+        $code->setCode($this->externalUserService->generateExternalUserCode());
+        $code->setTenant($user->getActiveTenant());
+        // Expire: 2 days
+        $code->setCodeExpire((new \DateTime())->add(new \DateInterval('P2D')));
```
turegjorup, line 50: Interval should be set as a class constant.

Nice to have: Interval can be set in `.env` instead

src/DataProvider/UserCollectionDataProvider.php
```diff
@@ -0,0 +1,51 @@
...
+        $currentRequest = $this->requestStack->getCurrentRequest();
+        $itemsPerPage = $currentRequest->query?->get('itemsPerPage') ?? 10;
+        $page = $currentRequest->query?->get('page') ?? 1;
+
+        /** @var User $currentUser */
+        $currentUser = $this->security->getUser();
+        $activeTenant = $currentUser->getActiveTenant();
+        $activeTenantUlid = $this->validationUtils->validateUlid($activeTenant->getId());
+
+        $queryBuilder = $this->userRepository->getExternalUsersByTenantQueryBuilder($activeTenantUlid);
```
turegjorup, line 40: It's misleading that a `UserCollectionDataProvider` only returns external users. At a minimum rename the class.

turegjorup: Not sure how to best handle this on the API level. on the one hand I would prefer one endpoint that can be filtered. On the other hand I'm not sure how clean an implementation that is in regards to access control. We would then on the security level have to enforce that the endpoint is called with the correct filter.

src/DataProvider/UserItemDataProvider.php
```diff
@@ -0,0 +1,40 @@
...
+    public function getItem(string $resourceClass, $id, string $operationName = null, array $context = []): ?User
+    {
+        $ulid = $this->validationUtils->validateUlid($id);
+
+        /** @var User $currentUser */
+        $currentUser = $this->security->getUser();
+        $activeTenant = $currentUser->getActiveTenant();
+        $activeTenantUlid = $this->validationUtils->validateUlid($activeTenant->getId());
+
+        return $this->userRepository->getExternalUserByIdAndTenant($ulid, $activeTenantUlid);
```
turegjorup, line 38: Same as for the CollectionProvider.Misleading that a `UserItemDataProvider` will only return an external user

src/Entity/User.php
```diff
@@ -17,10 +18,16 @@
...
 {
     /**
-     * @ORM\Column(type="string", length=180, unique=true)
+     * @ORM\Column(type="string", unique=true)
+     */
+    #[Assert\NotBlank]
+    private string $providerId = '';
+
+    /**
+     * @ORM\Column(type="string", length=180, nullable=true)
```
turegjorup, line 27: Even with a `$providerId` I still don't think `$email` should be nullable

turegjorup: And we should still enforce `unique=true` on emails

src/Entity/User.php
```diff
@@ -31,7 +38,7 @@ class User extends AbstractBaseEntity implements UserInterface, PasswordAuthenti
     /**
      * @var string The hashed password
      *
-     * @ORM\Column(type="string")
+     * @ORM\Column(type="string", nullable=true)
      */
     private string $password = '';
```
turegjorup, line 43: Either let it be nullable and change the default value to `null` or leave as is, non-nullable and empty string

src/Entity/User.php
```diff
@@ -45,6 +52,11 @@ class User extends AbstractBaseEntity implements UserInterface, PasswordAuthenti
      */
     private ?string $provider = null;
 
+    /**
+     * @ORM\Column(type="string", nullable=true, enumType="App\Enum\UserTypeEnum")
+     */
+    private ?UserTypeEnum $userType = null;
```
turegjorup, line 58: Why is this nullable? Should all users not have a type?

src/Repository/UserRepository.php
```diff
@@ -49,32 +102,14 @@ public function loadUserByIdentifier(string $identifier): ?User
             ->getOneOrNullResult();
     }
 
-    // /**
-    //  * @return User[] Returns an array of User objects
-    //  */
-    /*
-    public function findByExampleField($value)
+    public function getById(Ulid $ulid): Querybuilder
```
turegjorup, line 105: Does this not duplicate doctrine's own `find()` https://www.doctrine-project.org/projects/doctrine-orm/en/2.16/reference/working-with-objects.html#by-primary-key

src/Security/ExternalUserAuthenticator.php
```diff
@@ -0,0 +1,35 @@
...
+class ExternalUserAuthenticator extends JWTAuthenticator
+{
+    /** {@inheritDoc} */
+    final public function doAuthenticate(Request $request): Passport
+    {
+        $passport = parent::doAuthenticate($request);
+        $user = $passport->getUser();
+
+        if (!$user instanceof User) {
+            throw new AuthenticationException('UserInterface not of User type');
```
turegjorup, line 26: I have no idea what that exception message means

src/Service/ExternalUserService.php
```diff
@@ -0,0 +1,164 @@
...
+        $this->entityManager->persist($userRoleTenant);
+
+        $this->entityManager->flush();
+    }
+
+    public function getEmailFromDisplayName(string $displayName): string
+    {
+        $slugged = $this->slugifyDisplayName($displayName);
+
+        return "$slugged@ext";
```
turegjorup: I prefer `$slugged.'@ext'` or a `sprintf()` for readbility

src/Security/AzureOidcAuthenticator.php
```diff
@@ -47,27 +59,53 @@ public function authenticate(Request $request): Passport
...
+                    $signInName = $claims[$this->oidcExternalClaimId];
+
+                    if (empty($signInName)) {
+                        throw new CustomUserMessageAuthenticationException('Missing required claim '.$this->oidcExternalClaimId);
+                    }
+
+                    $type = UserTypeEnum::OIDC_EXTERNAL;
+                    $name = ExternalUserService::EXTERNAL_USER_DEFAULT_NAME;
+                    $providerId = $this->externalUserService->generatePersonalIdentifierHash($signInName);
+                    $email = null;
```
turegjorup, line 75: Not sure what the best solution is if the email is not nullable. Either do `$signInName.'@ext'` as a temporary value or make email nullable, but unique. (MySQL allows multiple NULLs in a column with a unique constraint.)

src/Service/ExternalUserService.php
```diff
@@ -0,0 +1,164 @@
...
+        return hash('sha512', $this->hashSalt.$personalIdentifier);
+    }
+
+    /**
+     * @throws CodeGenerationException
+     */
+    public function refreshCode(UserActivationCode $code): UserActivationCode
+    {
+        $code->setCode($this->generateExternalUserCode());
+        $code->setCodeExpire((new \DateTime())->add(new \DateInterval('P2D')));
```
turegjorup: We should not have this hardcoded two different places (Also found in `ExternalUserActivationCodeInputDataTransformer`). Promote it to a public class constant here, and then reference it. 

(Or better, make it and `.env` value)

src/DataTransformer/ExternalUserActivationCodeInputDataTransformer.php
turegjorup: Update: See comment for `ExternalUserService`

src/Service/ExternalUserService.php
```diff
@@ -0,0 +1,164 @@
...
+            ++$i;
+        } while ($i < 100);
+
+        throw new CodeGenerationException('Could not generate unique code.');
+    }
+
+    private function generateRandomCode(): string
+    {
+        $length = 12;
+        $chars = '0123456789ABCDEFGHIJKLMNOPQRSTUVWXYZ';
```
turegjorup: Promote to class constant

turegjorup (changes requested):

src/Repository/UserRepository.php
tuj: I think we need this because of 
->setParameter('user', $ulid, 'ulid');
'ulid' as type

src/Entity/User.php
tuj: Fixed

src/Service/ExternalUserService.php
tuj: Fixed

src/Security/AzureOidcAuthenticator.php
tuj: Will create a temporary email from providerId when the user has not activated. The mail will be updated after activation.

src/Service/ExternalUserService.php
tuj: Fixed

src/Entity/User.php
tuj: Changed back

src/Entity/User.php
tuj: Fixed

config/api_platform/user-activation-code.yaml
```diff
@@ -0,0 +1,69 @@
+---
+resources:
+  App\Entity\Tenant\UserActivationCode:
+    itemOperations:
+      get:
+        security: 'is_granted("ROLE_ADMIN") or is_granted("ROLE_EXTERNAL_USER_ADMIN")'
+      delete:
+        security: 'is_granted("ROLE_ADMIN") or is_granted("ROLE_EXTERNAL_USER_ADMIN")'
+      refresh_code:
+        security: 'is_granted("ROLE_ADMIN") or is_granted("ROLE_EXTERNAL_USER_ADMIN")'
```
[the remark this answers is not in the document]
tuj: Fixed

config/api_platform/user-activation-code.yaml
```diff
@@ -0,0 +1,69 @@
...
+        method: 'POST'
+        path: '/user-activation-codes/{id}/refresh-code'
```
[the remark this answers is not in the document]
tuj: I think I will leave it as a separate endpoint since it relates to an entity

config/api_platform/user.yaml
```diff
@@ -0,0 +1,90 @@
+---
+resources:
+  App\Entity\User:
+    attributes:
+      security: 'is_granted("ROLE_ADMIN") or is_granted("ROLE_EXTERNAL_USER_ADMIN")'
```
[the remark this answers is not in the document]
tuj: fixed

src/DataTransformer/ExternalUserActivationCodeInputDataTransformer.php
tuj: Fixed

src/DataProvider/UserCollectionDataProvider.php
tuj: I have removed the external part.
I have made a doctrine extension to limit access.

src/DataProvider/UserItemDataProvider.php
tuj: Fixed

src/Service/ExternalUserService.php
tuj: Fixed

src/Controller/UserActivateController.php
tuj: Fixed

src/Controller/UserRefreshCodeController.php
tuj: Fixed

src/Controller/UserRefreshCodeController.php
tuj: Fixed

src/Security/ExternalUserAuthenticator.php
tuj: Fixed

PR #220, opened 2024-10-24
Feature/bug screen playlist region no coding standards

tuj:


turegjorup (approved):
Approved with commit:
* Validate input object to ensure we have needed array keys
* Clarified parameter naming
* Ensure proper null checks
* Include id in error messages if content not found to ease debugging for clients

PR #236, opened 2025-03-19
Notified Feed Type bug when media url is null

tuj:
#### Link to ticket

https://leantime.itkdev.dk/#/tickets/showTicket/3941

#### Description

Fixed bug where no media url made Notified feed crash.

#### Checklist

### DIFF
--- a/src/Feed/NotifiedFeedType.php
+++ b/src/Feed/NotifiedFeedType.php
@@ -57,11 +57,15 @@ class NotifiedFeedType implements FeedTypeInterface
             // Check that image is available and accessible, otherwise leave out the feed element.
             foreach ($feedItems as $feedItem) {
                 if (!empty($feedItem['mediaUrl'])) {
-                    $response = $this->client->request(Request::METHOD_HEAD, $feedItem['mediaUrl']);
-                    $statusCode = $response->getStatusCode();
+                    try {
+                        $response = $this->client->request(Request::METHOD_HEAD, $feedItem['mediaUrl']);
+                        $statusCode = $response->getStatusCode();
 
-                    if (200 == $statusCode) {
-                        $result[] = $feedItem;
+                        if (200 === $statusCode) {
+                            $result[] = $feedItem;
+                        }
+                    } catch (\Exception) {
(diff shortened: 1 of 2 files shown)


src/Feed/NotifiedFeedType.php
```diff
@@ -54,13 +54,15 @@ public function getData(Feed $feed): array
...
             foreach ($feedItems as $feedItem) {
-                $response = $this->client->request(Request::METHOD_HEAD, $feedItem['mediaUrl']);
-                $statusCode = $response->getStatusCode();
+                if (!empty($feedItem['mediaUrl'])) {
+                    $response = $this->client->request(Request::METHOD_HEAD, $feedItem['mediaUrl']);
+                    $statusCode = $response->getStatusCode();
 
-                if (200 == $statusCode) {
-                    $result[] = $feedItem;
+                    if (200 == $statusCode) {
```
turegjorup, line 63:
```suggestion
                    if (200 === $statusCode) {
```

src/Feed/NotifiedFeedType.php
```diff
@@ -54,13 +54,15 @@ public function getData(Feed $feed): array
 
             $result = [];
 
-            // Check that image is accessible, otherwise leave out the feed element.
+            // Check that image is available and accessible, otherwise leave out the feed element.
             foreach ($feedItems as $feedItem) {
-                $response = $this->client->request(Request::METHOD_HEAD, $feedItem['mediaUrl']);
-                $statusCode = $response->getStatusCode();
+                if (!empty($feedItem['mediaUrl'])) {
+                    $response = $this->client->request(Request::METHOD_HEAD, $feedItem['mediaUrl']);
```
turegjorup, line 60: We should wrap http request in a try/catch, nor all failures will be covered by checking for a `200` status code. Not sure we need to do anything on failure execept exclude the `$feedItem` from `$result` as is done for non-200 status codes. (We could log it if we have a logger for the feed types, but I doubt it would have any practical use)

turegjorup (changes requested):

turegjorup (approved):

PR #242, opened 2025-05-12
Changed calendar api feed type to use display name from resources feed instead of events feed

tuj:
#### Link to ticket

https://leantime.itkdev.dk/#/tickets/showTicket/4423

#### Description

Changed calendar api feed type to use display name from resources feed instead of events feed.

#### Checklist

### DIFF
--- a/src/Feed/CalendarApiFeedType.php
+++ b/src/Feed/CalendarApiFeedType.php
@@ -79,7 +79,7 @@ class CalendarApiFeedType implements FeedTypeInterface
             $events = [];
 
             foreach ($requestedResources as $requestedResource) {
-                $events += $this->getResourceEvents($requestedResource);
+                $events = array_merge($events, $this->getResourceEvents($requestedResource));
             }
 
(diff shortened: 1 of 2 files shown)


src/Feed/CalendarApiFeedType.php
```diff
@@ -72,24 +72,35 @@ public function getData(Feed $feed): array
...
+            $requestedResources = $configuration['resources'];
+
+            $allResources = $this->loadResources();
 
             $events = [];
 
-            foreach ($resources as $resource) {
-                $events += $this->getResourceEvents($resource);
+            foreach ($requestedResources as $requestedResource) {
+                $events += $this->getResourceEvents($requestedResource);
```
turegjorup, line 82:
```suggestion
                $events[] = $this->getResourceEvents($requestedResource);
```

Personal preference but I like this syntax better for arrays. It help communicate the (array) data type

turegjorup (approved):

src/Feed/CalendarApiFeedType.php
tuj: Changed to array_merge

Issue #249, opened 2025-06-25
Release 3.0.0


tuj:
#### Description

This release is a "maintenance" release that improves the developer experience and cleans up technical debt.

#### Linked Issues

* https://github.com/os2display/display-api-service/issues/247
* https://github.com/os2display/display-api-service/issues/248

#### TODOs

- [x] Merge the code repositories into the API repository
- [x] Remove remote components
- [x] Better config handling through Symfony
- [x] Fix hmr setup
- [x] Release.json
- [x] Unify changelogs
- [x] Setup admin tests with Playwright.
- [ ] Setup client tests with Playwright
- [ ] Setup template tests with Playwright
- [ ] Developer docs
- [ ] Traefik-less dev setup
- [ ] Update docs with client configuration values
- [ ] GitHub actions for all parts of the code
- [ ] Cleanup infrastructure
- [ ] Cleanup docker image builds
- [ ] .docker conf files - make sure this is handled correctly
- [ ] Update documentation
- [ ] Write upgrade guide 2.x -> 3.0
- [ ] Setup latest github actions from template
- [ ] Simplify template install
- [ ] Add ADR
- [ ] Fix vite setup to make tests runnable in dev mode
- [ ] Update with recent code changes
- [ ] Integrate slide forms in admin instead of loading json files
- [ ] Replace quill (with https://tiptap.dev/docs ?)
- [ ] Replace lodash.set
- [ ] Cleanup frontend logging (pino). Is not configurable atm.
- [ ] Write how custom templates can be added. Fork.
- [ ] Update of api spec should update redux api slices as well.
- [ ] Upgrade all dependencies

#### Breaking changes

* External templates cannot be loaded anymore, since remote-components has been removed.
* All config.json files are not used anymore. All configuration should be set through .env
* All paths to the API are hardcoded in the client and admin.
* The clients use the same release.json file.
* Js tests should be run on production build files.

agnetemoos:
I really like the to-do list—great idea!
I’ll keep an eye on it and add to it if I come across anything.

PR #271, opened 2025-08-31
Online check

tuj:
#### Link to issue

#### Description

Added online check.

#### Checklist


CHANGELOG.md
```diff
@@ -15,6 +15,7 @@ All notable changes to this project will be documented in this file.
 * Upgraded redux-toolkit and how api slices are generated.
 * Fixed redux-toolkit cache handling.
 * Add Taskfile
+* Added online-check to public.
```
turegjorup, line 18:
```suggestion
* Added (screen) client online-check to public.
```

turegjorup (changes requested):
The current URL is `https://os2display.example.com/client/online-check/` (note `/client`). This will change that to `https://os2display.example.com/online-check/` without making a redirect which is an unnecessary breaking change.

Please move the files to `/public/client/online-check` so that 1) we don't change the url, 2) it scopes the files to the client

PR #272, opened 2025-08-31
Update command

tuj:
#### Link to issue

#### Description

Added update command.

#### Checklist

### DIFF
--- a/src/Command/UpdateCommand.php
+++ b/src/Command/UpdateCommand.php
@@ -10,7 +10,7 @@ use Symfony\Component\Console\Command\Command;
 use Symfony\Component\Console\Input\ArrayInput;
 use Symfony\Component\Console\Input\InputInterface;
 use Symfony\Component\Console\Output\OutputInterface;
-use Symfony\Component\Console\Question\Question;
+use Symfony\Component\Console\Question\ConfirmationQuestion;
 use Symfony\Component\Console\Style\SymfonyStyle;
 
 #[AsCommand(
@@ -57,8 +57,7 @@ class UpdateCommand extends Command
 
         // If no installed templates, we assume that this is a new installation and offer to install all templates.
         if ($isInteractive && 0 === count($installedTemplates)) {
-            $question = new Question('No templates are installed. Install all '.count($allTemplates).'?', 'yes');
-            $question->setAutocompleterValues(['yes', 'no']);
+            $question = new ConfirmationQuestion('No templates are installed. Install all '.count($allTemplates).'?');
             $installAll = $io->askQuestion($question);
 
             if ('yes' === $installAll) {


src/Command/UpdateCommand.php
```diff
@@ -4,9 +4,89 @@
...
+    description: 'Check if important commands have been run. Use --force to run the commands.',
+)]
+class UpdateCommand extends Command
 {
-    // TODO: Test that migrations have been run.
-    // TODO: Run test of status for templates. No templates = clean install. Install all?
-    // TODO: Update existing templates.
+    protected function configure(): void
+    {
+        $this->addOption('force', 'f', InputOption::VALUE_NONE, 'Update all requirements. Otherwise, update command only checks requirements.');
```
turegjorup, line 23: I don't like the naming combination here. `--force` means something else for me. If we need a status command I would make a dedicated `app:status`.

For `aap:update` we could follow doctrine migrations:
* no option: Prompt for "are yoy sure?"
* `--no-interaction`: just run

src/Command/UpdateCommand.php
```diff
@@ -4,9 +4,89 @@
...
+            $application->doRun($migrationsCommand, $output);
+
+            // Install all templates.
+            $migrationsCommand = new ArrayInput([
+                'command' => 'app:templates:install',
+                '--all' => true,
+                '--update' => true,
+            ]);
+            $migrationsCommand->setInteractive(false);
+            $application->doRun($migrationsCommand, $output);
```
turegjorup, line 55: If the called command fail we should also fail.

turegjorup (changes requested):
I think we need to iterate a little on the `app:update` command. I would expect it to just update everything (migrations and templates). How many confirmations we prompt for i don't know. Could be just at first, could be individually for migrations and for templates.

See https://github.com/kimai/kimai/blob/main/src/Command/InstallCommand.php for inspiration

src/Command/UpdateCommand.php
```diff
@@ -4,9 +4,79 @@
...
+            return Command::FAILURE;
+        }
+
+        $allTemplates = $this->templateService->getAllTemplates();
+        $installedTemplates = array_filter($allTemplates, fn ($entry): bool => $entry->installed);
+
+        // If no installed templates, we assume that this is a new installation and offer to install all templates.
+        if ($isInteractive && 0 === count($installedTemplates)) {
+            $question = new Question('No templates are installed. Install all '.count($allTemplates).'?', 'yes');
+            $question->setAutocompleterValues(['yes', 'no']);
```
turegjorup, line 61: Symfony has a `ConfirmationQuestion` https://symfony.com/doc/current/components/console/helpers/questionhelper.html#asking-the-user-for-confirmation

turegjorup (approved):
Approved with one minor suggestion

PR #288, opened 2025-09-06
Added command to migrate config.json files to .env variables

tuj:
#### Issue

#### Description

- Added command to migrate config.json files to .env variables.
- Expanded upgrade guide.

#### Checklist


turegjorup (changes requested):
Some suggestions in comments.

Most important is the issue of default values. The convention for `.env[.*]` in Symfony applications is that default values are defined in `.env` and overrides in `.env.local`. 

This command defines defaults and asks that you set them in `.env.local`. It would be better to only give out put for lines with values set. If no value is set, the variable should be omitted to allow fallback to defaults in `.env`

turegjorup (approved):
This should potentially target the 2.* branch so that you can run it prior to upgrading to 3.0

PR #296, opened 2025-09-10
Screen layout commands and expanded documentation

tuj:
#### Issue

#### Description

- Added screen layout commands.
- Expanded documentation.

#### Checklist


turegjorup (approved):
This _can_ be merged but I think it would be beneficial to work more cleanly with available vs installed templates and layouts.

Logic for checking if a template or layout is installed is spread througout leading to code that is harder to follow and debug than necessary. I think it should be possible to consolidate that logic in the two service classes. Thus all other code could call `->getInstalledTemplates|Layouts()`. This way other code would be more readable .

turegjorup (approved):

PR #313, opened 2025-09-18
BRND Booking

agnetemoos:
#### Link to ticket

Fixes os2display/display-api-service#309

#### Description

Add BRND Booking feed type

### DIFF
--- a/src/Feed/BrndFeedType.php
+++ b/src/Feed/BrndFeedType.php
@@ -84,7 +84,7 @@ class BrndFeedType implements FeedTypeInterface
                 return $carry;
             }, []);
         } catch (\Throwable $throwable) {
-          $this->logger->error($throwable->getMessage());
+            $this->logger->error($throwable->getMessage());
             // Silently catch all exceptions and return empty result
             // $result is already initialized with empty bookings array
         }


src/Feed/SourceType/Brnd/ApiClient.php
```diff
@@ -0,0 +1,195 @@
...
+                $contentDecoded = json_decode($content, false, 512, JSON_THROW_ON_ERROR);
+                $token = $contentDecoded->data->access_token;
+
+                // Expire cache 5 min before token expire
+                $expireSeconds = intval(self::TOKEN_TTL - 300);
+                $cacheItem->set($token);
+                $cacheItem->expiresAfter($expireSeconds);
+                $this->feedsCache->save($cacheItem);
+            } catch (\Throwable $throwable) {
+                throw new BrndException($throwable->getMessage(), (int) $throwable->getCode(), $throwable);
```
tuj: If you can throw some of the exceptions from src/Exceptions: https://github.com/os2display/display-api-service/tree/develop/src/Exceptions they will be mapped by Api Platform. See https://github.com/os2display/display-api-service/blob/develop/config/packages/api_platform.yaml#L55

src/Feed/BrndFeedType.php
```diff
@@ -0,0 +1,183 @@
...
+                'type' => 'string',
+                'exposeValue' => true,
+            ],
+            'company_id' => [
+                'type' => 'string',
+                'exposeValue' => true,
+            ],
+            'api_auth_key' => [
+                'type' => 'string',
+                'exposeValue' => true,
```
tuj: I think you should hide this value, otherwise everybody in the admin can see the value.

```suggestion
                'exposeValue' => false,
```

It will still be possible to set the value, it will just not be visible in the admin after being set.

src/Feed/BrndFeedType.php
```diff
@@ -0,0 +1,183 @@
...
+        return $result;
+    }
+
+    private function parseBrndBooking(array $booking): array
+    {
+        // Parse start time
+        $startDateTime = null;
+        if (!empty($booking['dato']) && isset($booking['starttid']) && is_string($booking['starttid'])) {
+            // Trim starttid to 6 digits after dot for microseconds
+            $starttid = preg_replace('/\.(\d{6})\d+$/', '.$1', $booking['starttid']);
```
tuj: Replace danish variable name with english.
Maybe: $startTimeString

tuj (changes requested):
Generally, looks good.
I have a few issues I think you should handle.

src/Feed/BrndFeedType.php
agnetemoos: The danish varible names are in the response object returned from the BRND webservice. They have choosen to code their webservice in danish. An odd choice, but I can not change that. I have attached an example of a webservice response so you can see it for yourself.

agnetemoos: [response_1761723886623.json](https://github.com/user-attachments/files/23233436/response_1761723886623.json)

src/Feed/BrndFeedType.php
```diff
@@ -0,0 +1,183 @@
...
+
+        $feedSource = $feed->getFeedSource();
+
+        if (null === $feedSource) {
+            return $result;
+        }
+
+        $bookings = $this->apiClient->getInfomonitorBookingsDetails($feedSource, $sportCenterId);
+
+        $result['bookings'] = array_map([$this, 'parseBrndBooking'], $bookings);
```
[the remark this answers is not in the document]
agnetemoos: I have added a few requrements, but I prefer to expect very little. BRND is a sprawling system with a lot of use cases unknown to me.

tuj (changes requested):
Looks good.
Added suggestion for a log message. After this is added it will be approved.

tuj (approved):

PR #323, opened 2025-11-03
update infrastructure for 3.0

turegjorup:
#### Link to ticket

#### Description

Consolidates the four parallel image builds (`itkdev/*` × `os2display/*` × `display-api-service` + `nginx`) into two GHCR-published images and rewires CI accordingly.

**Image build pipeline**
- Two images at `ghcr.io/os2display/display-api-service` and `ghcr.io/os2display/display-api-service-nginx` (parallel `itkdev/*` and `os2display/*` trees deleted).
- New `.github/workflows/build-images.yml` triggers on `develop` push and tag push; uses `docker/metadata-action` with `latest=auto` so `3.0.0` updates `:latest` and `3.0.0-RC1` doesn't.
- Multi-arch (`linux/amd64`, `linux/arm64`) with `provenance: mode=max` and `sbom: true`.
- Nginx image layers on top of the just-published API image as a `FROM` stage rather than rebuilding the npm/composer stages — `npm ci` + `composer install` run once per CI run instead of twice. Local nginx build resolved in ~1s after API was loaded; nginx image size 108 MB (vs. 170 MB on the previous design).
- Old `infrastructure/build-n-push.sh` (private namespace, single-platform `--push`) replaced with multi-arch `buildx`; `BUILD_LOAD=1` builds host-platform-only and loads to local daemon for smoke tests.

**Image internals**
- Nginx config consolidates the per-fork `confd`-templated configs into the upstream `nginxinc/nginx-unprivileged` template mechanism (`infrastructure/nginx/etc/templates/default.conf.template`) with explicit `NGINX_FPM_SERVICE` / `NGINX_FPM_PORT` / `NGINX_MAX_BODY_SIZE` / `NGINX_SET_REAL_IP_FROM` / `NGINX_PORT` env contract.
- API image switches to `ENTRYPOINT ["docker-entrypoint.sh"] / CMD ["php-fpm"]` so the `composer dump-env prod` + `cache:warmup` preamble runs for any command, not just the default.
- Entrypoint runs with `set -eu` (no `-x`); composer-1 conditional removed (image only ships composer 2).
- Multi-line `ENV` block in the API Dockerfile collapsed to a single declaration (blank lines between `\` continuations are not portable).
- `APP_RELEASE_VERSION` build arg dropped (declared but never referenced).
- Nginx `/health` location is exact-match (`location =`) instead of prefix.

**dotenv / Symfony config surface**
- Symfony `.env` extended to be the documented source of truth for every variable read via `%env(...)%` in `config/`. Each variable carries a description comment.
- New `scripts/check-env-coverage.sh` enforces both directions in CI via `.github/workflows/env_coverage.yaml`: every `%env(VAR)%` consumed in `config/` must be declared in `.env`, and every variable in `.env` must have a preceding description comment.
- `config/packages/pentatrion_vite.yaml` added.

**Compose / dev workflow**
- `docker-compose.yml`, `docker-compose.server.yml`, `docker-compose.override.yml` updated for the new `phpfpm` + `nginx` topology and for the `os2display` ↔ `display` project rename.
- Nginx healthcheck wired in dev (`/health`) so Playwright gates on it.
- `Taskfile.yml` gains `coding-standards:env:check`.

**CI workflow cleanup**
- `apispec.yaml`, `phpunit.yaml`, `code-analysis.yaml`, `composer_install.yaml`, `doctrine.yaml`, `rector.yaml` simplified to the same `docker compose run --rm phpfpm …` shape.
- `actions/checkout@v6` everywhere; `markdown.yaml`, `styles.yaml`, `twig.yaml`, `vitest.yaml`, `yaml.yaml`, `composer.yaml`, `javascript.yaml`, `php.yaml` aligned.
- `playwright.yaml` gates on the nginx healthcheck.
- Old per-fork `itkdev_docker_build_*.yml` and `os2display_docker_build_*.yml` workflows deleted.
- `github_build_release.yml` writes `release.json` into the tarball and emits pre-releases for tags containing `-` (mirrors `latest=auto`).
- `.git-blame-ignore-revs` added so the bulk style commit doesn't pollute `git blame`.

**`.dockerignore`**
- New file; excludes dev tooling, IDE state, CI configs, vendor/node_modules from the build context.

#### Smoke test

`BUILD_LOAD=1 sh infrastructure/build-n-push.sh` builds both images locally end-to-end. Verified the nginx image carries `/var/www/html/public/release.json` and `public/build/.vite/`, the API container exposes the full `PHP_OPCACHE_*` / `PHP_PM_*` env surface, and the API entrypoint runs `set -eu`.

#### Screenshot of the result

N/A

#### Checklist

Two follow-ups intentionally deferred to subsequent PRs (per the consolidation plan):
- Single ISO-formatted `releaseTime` and one-place `release.json` write.
- Shipping `.env.production.example` in the API image so operators can bootstrap `.env.local` via `docker run --rm <image> cat …`.


infrastructure/display-api-service/docker-entrypoint.sh
```diff
@@ -2,10 +2,8 @@
 
 set -eux
 
-## Run templates with configuration.
-/usr/local/bin/confd --onetime --backend env --confdir /etc/confd
-
-## Bump env.local into PHP for better performance.
+## Dump dotenv files into PHP for better performance.
+## @see https://symfony.com/doc/6.4/configuration.html#configuring-environment-variables-in-production
 composer dump-env prod
```
tuj, line 7: This can be annoying in production, when you want to change an environment variable.

infrastructure/nginx/Dockerfile
```diff
@@ -0,0 +1,42 @@
...
+# Pull the API image that was built and pushed in the preceding step.
+# Multi-arch manifests resolve per platform, so the arm64 nginx build
+# copies arm64 contents and amd64 copies amd64 — no QEMU re-execution
+# of npm/composer for the second image.
+FROM ${APP_IMAGE}:${APP_VERSION} AS app
+
+
+######## Nginx production image ########
+FROM nginxinc/nginx-unprivileged:alpine
+LABEL maintainer="ITK Dev <itkdev@mkb.aarhus.dk>"
```
tuj: correct maintainer email?

infrastructure/display-api-service/Dockerfile
```diff
@@ -0,0 +1,109 @@
+######## Clients [Screen|Admin] build ########
+FROM node:24-alpine AS client_app_builder
+LABEL maintainer="ITK Dev <itkdev@mkb.aarhus.dk>"
```
tuj: correct email?

infrastructure/display-api-service/Dockerfile
```diff
@@ -0,0 +1,109 @@
...
+COPY --from=repository-root assets/ ./assets/
+COPY --from=repository-root public/client/ ./public/client/
+
+# Build the application with version info
+RUN npm run build
+
+
+######### API backend build ########
+FROM itkdev/php8.4-fpm:alpine AS api_app_builder
+LABEL maintainer="ITK Dev <itkdev@mkb.aarhus.dk>"
```
tuj: correct email?

infrastructure/display-api-service/Dockerfile
```diff
@@ -0,0 +1,109 @@
...
+RUN printf '{"releaseTimestamp": %s, "releaseTime": "%s", "releaseVersion": "%s"}\n' \
+      "${APP_RELEASE_TIMESTAMP}" \
+      "${APP_RELEASE_TIME}" \
+      "${APP_VERSION}" \
+      > public/release.json
+
+
+######## PHP-FPM (API) production image ########
+FROM itkdev/php8.4-fpm:alpine
+LABEL maintainer="ITK Dev <itkdev@mkb.aarhus.dk>"
```
tuj: correct email?

infrastructure/nginx/Dockerfile
```diff
@@ -0,0 +1,42 @@
...
+    NGINX_FPM_PORT=9000 \
+    NGINX_MAX_BODY_SIZE=140m \
+    NGINX_SET_REAL_IP_FROM=172.16.0.0/12 \
+    NGINX_WEB_ROOT=/var/www/html/public
+
+WORKDIR ${APP_API_PATH}
+
+# Single source of truth: public/ (incl. build/ and release.json) comes
+# from the just-published API image, not a duplicated builder stage.
+COPY --from=app --chown=$UID:0 --chmod=775 ${APP_API_PATH}/public ./public
```
tuj: Can we use workdir instead of /var/www/html ?

infrastructure/nginx/Dockerfile
```diff
@@ -0,0 +1,42 @@
...
+
+ARG UID=101
+
+ENV APP_API_PATH=/var/www/html \
+    NGINX_PORT=8080 \
+    NGINX_FPM_SERVICE=os2display \
+    NGINX_FPM_PORT=9000 \
+    NGINX_MAX_BODY_SIZE=140m \
+    NGINX_SET_REAL_IP_FROM=172.16.0.0/12 \
+    NGINX_WEB_ROOT=/var/www/html/public
```
tuj, line 24: Are all values used?

infrastructure/build.sh
```diff
@@ -0,0 +1,52 @@
+#!/bin/sh
```
tuj, line 1: Can we split build and push, so it is not done accidentially?

.git-blame-ignore-revs
```diff
@@ -0,0 +1,9 @@
+# Revisions to skip when running `git blame`.
```
tuj, line 1: What is this?

.dockerignore
```diff
@@ -0,0 +1,73 @@
+/.git
+!/.git/config
+
+# Unneeded application files
+/.docker
+/.github
+/docs
+/fixtures
+/infrastructure
+/public/build
```
tuj, line 10: Should this be ignored?

.dockerignore
```diff
@@ -0,0 +1,73 @@
...
+/psalm-baseline.xml
+/rector.php
+
+###> symfony/framework-bundle ###
+/.env.local
+/.env.local.php
+/.env.*.local
+/config/secrets/prod/prod.decrypt.private.php
+/var/
+/vendor/
```
tuj, line 40: Should this be ignored?

infrastructure/display-api-service/Dockerfile
```diff
@@ -0,0 +1,109 @@
+######## Clients [Screen|Admin] build ########
```
tuj: I think we should work through this file and see if some things can simplified.

reference: https://github.com/itk-dev/os2display-api-service/blob/0b2cc18e2dd5a637f14e6df0bd86a379af37b3a8/infrastructure/display-api-service/Dockerfile

infrastructure/display-api-service/docker-entrypoint.sh
tuj: Can we avoid this file?

tuj (changes requested):

infrastructure/display-api-service/docker-entrypoint.sh
turegjorup: We dump it on container start, so it will always be up to date, there is no manual step. Images are intended for production use, not for testing/debugging so I consider it an edge case.

Have made note to ensure it is properly documented .

.git-blame-ignore-revs
turegjorup: Major style change commits can be noisy in git blame. Mostly when running git blame we want to answer the question "who implemented this feature" not "who ran the last code style update". This is the git/github standard way of solving it: https://docs.github.com/en/repositories/working-with-files/using-files/viewing-and-understanding-files#ignore-commits-in-the-blame-view

.dockerignore
turegjorup: Yes. Vite generates it during the API image build (`api_app_builder` stage line 56). Including the host's local `public/build` would bake stale or `dev-mode` assets into the image.

.dockerignore
turegjorup: Yes. composer install runs in `api_app_builder` stage line 48/59 to populate vendor/ deterministically inside the image. Letting the host's vendor/ leak into the build context would make image content depend on the developer's local state.

infrastructure/display-api-service/docker-entrypoint.sh
turegjorup: Also values set as "proper" environment values for the container will always take precedence over values in `.env.*` files. So any values passed through `env_file` or `environment` in docker compose or other container runtime will override anything in set `.env`.

infrastructure/build.sh
turegjorup: Done. Also added as task that prompts for confirm on push

infrastructure/nginx/Dockerfile
turegjorup: Yes

infrastructure/nginx/Dockerfile
turegjorup: Refactored

infrastructure/display-api-service/Dockerfile
turegjorup: Done in `393cb3d0` after a side-by-side with the canonical reference.

**Adopted:**
- Stage rename `client_app_builder` → `assets_builder`
- Canonical section-divider style (`################################`)
- Dropped `APP_CLIENT_PATH` / `APP_API_PATH` env indirections — literal paths throughout
- Dropped the dead `APP_RELEASE_TIMESTAMP` arg in the assets stage
- Expanded inline comments to document the *why* (two-pass composer install, vite manifest ordering ahead of `cache:clear`, runtime composer for the entrypoint, Prometheus exporter, ENTRYPOINT vs CMD)

**Deliberately not adopted:**
- `--platform=$BUILDPLATFORM` on the final stage — multi-arch correctness bug. `BUILDPLATFORM` is the host doing the build; pinning the final stage to it publishes build-host binaries in *every* manifest entry (the `linux/arm64` entry from an amd64 runner becomes silently amd64).
- `itkdev/php8.4-fpm:latest` for builder + `:alpine` for final — researched the tag structure: they're parallel variants (`:latest` tracks `2.x.y` Ubuntu, `:alpine` tracks `alpine-1.4.x` Alpine), not version aliases. Both stages stay on `:alpine` for consistency; added a TODO to pin to a specific `alpine-1.4.x` (or digest) once we settle on a baseline so builds become reproducible.
- `RUN rm -rf package* vite.config.js` cleanup — only affects the discarded builder layer; the final image only `COPY`s the `vendor/`+source tree. Not worth the noise.
- Single-pass `COPY . ./` then `composer install` — kept the two-pass pattern (lockfile-only `--no-scripts` install, then full source + scripted install) so the `vendor/` layer caches across pure source-code edits.

turegjorup: Have adopted some things from the linked file. Summary of adopted/not-adopted below

**Adopted:**
- Stage rename `client_app_builder` → `assets_builder`
- Canonical section-divider style (`################################`)
- Dropped `APP_CLIENT_PATH` / `APP_API_PATH` env indirections — literal paths throughout
- Dropped the dead `APP_RELEASE_TIMESTAMP` arg in the assets stage
- Expanded inline comments to document the *why* (two-pass composer install, vite manifest ordering ahead of `cache:clear`, runtime composer for the entrypoint, Prometheus exporter, ENTRYPOINT vs CMD)

**Deliberately not adopted:**
- `--platform=$BUILDPLATFORM` on the final stage — multi-arch correctness bug. `BUILDPLATFORM` is the host doing the build; pinning the final stage to it publishes build-host binaries in *every* manifest entry (the `linux/arm64` entry from an amd64 runner becomes silently amd64).
- `itkdev/php8.4-fpm:latest` for builder + `:alpine` for final — researched the tag structure: they're parallel variants (`:latest` tracks `2.x.y` Ubuntu, `:alpine` tracks `alpine-1.4.x` Alpine), not version aliases. Both stages stay on `:alpine` for consistency; added a TODO to pin to a specific `alpine-1.4.x` (or digest) once we settle on a baseline so builds become reproducible.
- `RUN rm -rf package* vite.config.js` cleanup — only affects the discarded builder layer; the final image only `COPY`s the `vendor/`+source tree. Not worth the noise.
- Single-pass `COPY . ./` then `composer install` — kept the two-pass pattern (lockfile-only `--no-scripts` install, then full source + scripted install) so the `vendor/` layer caches across pure source-code edits.

infrastructure/display-api-service/docker-entrypoint.sh
```diff
@@ -0,0 +1,12 @@
+#!/bin/sh
+
+set -eu
+
+## Dump dotenv files into PHP for better performance.
+## @see https://symfony.com/doc/6.4/configuration.html#configuring-environment-variables-in-production
+composer dump-env prod
```
tuj, line 7: Maybe we need an entry in README.md about the images, that links to infrastructure/Readme.md and explains what should be done if you want to change env variables with the images?

infrastructure/display-api-service/Dockerfile
```diff
@@ -0,0 +1,117 @@
...
+
+# Lockfile + vite config first so the npm install layer caches across
+# unrelated source edits.
+COPY --from=repository-root package.json package-lock.json vite.config.js ./
+RUN npm ci --no-audit --no-fund
+
+# Source for `npm run build`. Vite emits to /app/public/build, which the
+# api_app_builder stage copies out.
+COPY --from=repository-root assets/ ./assets/
+COPY --from=repository-root public/client/ ./public/client/
```
tuj: Add a note that this is to make sure client online-check exists in public

tuj: I think we also need: https://github.com/os2display/display-api-service/tree/release/3.0.0/public/media

infrastructure/display-api-service/Dockerfile
```diff
@@ -0,0 +1,117 @@
...
+COPY --chown=deploy:deploy --from=repository-root composer.json composer.lock symfony.lock ./
+RUN APP_ENV=prod composer install --no-dev -o --classmap-authoritative --no-scripts
+
+COPY --chown=deploy:deploy --from=repository-root ./ ./
+
+# Vite manifest must be in place before the second composer install
+# triggers cache:clear, otherwise the vite bundle can't generate its
+# cache config files.
+# @see https://symfony-vite.pentatrion.com/guide/performance.html#caching-configuration-files-%F0%9F%8F%83
+COPY --chown=deploy:deploy --from=assets_builder /app/public/build ./public/build
```
tuj, line 70: Can this step be moved before the first composer install to avoid running composer install two times?

tuj (changes requested):

infrastructure/display-api-service/Dockerfile
turegjorup: Separately: we need to document that uploads need to be writable and mapped to a persistent volume.

As for the two production-needed thumbnails — they're kept by `.dockerignore` lines 13-14:

```
!public/media/thumbnail_other.png
!public/media/thumbnail_video.png
```

They reach the final image via the api_app_builder's `COPY --from=repository-root ./ ./`.

Verified inside the built image:

> -rw-r--r--    1 deploy deploy   7426 /var/www/html/public/media/thumbnail_other.png
> -rw-r--r--    1 deploy deploy   8346 /var/www/html/public/media/thumbnail_video.png

infrastructure/display-api-service/Dockerfile
turegjorup: Doing double composer install is deliberate. It's done to optimize the image layer cache. First pass (`--no-scripts`, line 55) runs against `composer.lock` only, before any source code is copied. The resulting `vendor/` layer stays cached as long as the lockfile is unchanged, even if every source file changes.

The assumption is that application code changes more frequently than vendor code. If that is true then splitting vendor code and application code in two image layers that can be cached independently makes sense.

It's not a problem to structure the build file to only do one `composer install`, but then we lose the independent cache layers. Also note that the second `composer install` doesn't do a complete re-install given that the dependencies are already installed. It re-runs the autoloader optimisation against the now-present source and triggers Symfony's auto-scripts (`cache:clear`, `assets:install`).

tuj (approved):

PR #372, opened 2026-03-31
List optimizations

### LINKED ISSUE #369: V3: Elementer forsætter med spinne på skærmoversigt
Nogle elementer fx antal tilknyttede grupper bliver ved med at spinne på skærm oversigten hvor vi også kan se skærm status

Image

tuj:
#### Link to issue

https://github.com/os2display/display-api-service/issues/369

#### Link to ticket

https://leantime.itkdev.dk/#/tickets/showTicket/6871

#### Description

Atm. when lists fetch screens, groups, playlists several extra API calls are made to fetch how many relations are connected to each list element. 
Instead of this approach, this PR adds relations lengths to the API where it is needed and uses these lengths in the admin.

This PR also cleans up release.json loading to avoid fetching the file more often than every 5 minutes for both admin and client.

NB! This changes the API specification by adding optional length fields.

#### Screenshot

Screenshot 2026-04-07 at 10 06 31

#### Checklist


turegjorup:
Two alternative options using DQL, i prefer Option A over B and native SQL, but it's mostly personal preference. Would be nice to avoid (more) native SQL, but "nice" not "need"....

Option A: Two Simple Subqueries (most readable)

```php
  public function getCampaignCountsForScreen(Ulid $screenId): array
  {
      $em = $this->getEntityManager();
      $now = new \DateTime();

      // Subquery: all campaign IDs for this screen (direct + via groups)
      $directQb = $em->createQueryBuilder()
          ->select('IDENTITY(sc.campaign)')
          ->from(ScreenCampaign::class, 'sc')
          ->where('sc.screen = :screenId');

      $groupQb = $em->createQueryBuilder()
          ->select('IDENTITY(sgc.campaign)')
          ->from(ScreenGroupCampaign::class, 'sgc')
          ->innerJoin('sgc.screenGroup', 'sg')
          ->innerJoin('sg.screens', 's')
          ->where('s.id = :screenId');

      // Total count
      $total = (int) $em->createQueryBuilder()
          ->select('COUNT(DISTINCT p.id)')
          ->from(Playlist::class, 'p')
          ->where('p.id IN ('.$directQb->getDQL().') OR p.id IN ('.$groupQb->getDQL().')')
          ->setParameter('screenId', $screenId, 'ulid')
          ->getQuery()
          ->getSingleScalarResult();

      // Active count
      $activeQb = $em->createQueryBuilder()
          ->select('COUNT(DISTINCT p.id)')
          ->from(Playlist::class, 'p')
          ->where('p.id IN ('.$directQb->getDQL().') OR p.id IN ('.$groupQb->getDQL().')')
          ->andWhere('p.publishedFrom IS NULL OR p.publishedFrom < :now')
          ->andWhere('p.publishedTo IS NULL OR p.publishedTo > :now')
          ->setParameter('screenId', $screenId, 'ulid')
          ->setParameter('now', $now);

      $active = (int) $activeQb->getQuery()->getSingleScalarResult();

      return ['total' => $total, 'active' => $active];
  }
```
  Pros: Pure DQL, easy to read, uses entity relationships.
  Cons: Two queries instead of one.

  ---
  Option B: Single Query with Conditional Count via DQL
```php
  public function getCampaignCountsForScreen(Ulid $screenId): array
  {
      $em = $this->getEntityManager();
      $now = new \DateTime();

      $directDql = $em->createQueryBuilder()
          ->select('IDENTITY(sc.campaign)')
          ->from(ScreenCampaign::class, 'sc')
          ->where('sc.screen = :screenId')
          ->getDQL();

      $groupDql = $em->createQueryBuilder()
          ->select('IDENTITY(sgc.campaign)')
          ->from(ScreenGroupCampaign::class, 'sgc')
          ->innerJoin('sgc.screenGroup', 'sg')
          ->innerJoin('sg.screens', 's')
          ->where('s.id = :screenId')
          ->getDQL();

      $qb = $em->createQueryBuilder()
          ->select('COUNT(DISTINCT p.id) AS total')
          ->addSelect(
              'SUM(CASE WHEN (p.publishedFrom IS NULL OR p.publishedFrom < :now) '.
              'AND (p.publishedTo IS NULL OR p.publishedTo > :now) THEN 1 ELSE 0 END) AS active'
          )
          ->from(Playlist::class, 'p')
          ->where("p.id IN ({$directDql}) OR p.id IN ({$groupDql})")
          ->setParameter('screenId', $screenId, 'ulid')
          ->setParameter('now', $now);

      $result = $qb->getQuery()->getSingleResult();

      return [
          'total' => (int) $result['total'],
          'active' => (int) $result['active'],
      ];
  }

  Pros: Single query, closest to the original native SQL behavior.
  Cons: The SUM(CASE...) in DQL is less readable; may not work with all Doctrine versions (CASE in SELECT requires Doctrine ORM 2.6+).
```

turegjorup (changes requested):
Any `$object->getFooBarGroups()->count();` call still loads all objects in the relation. This is basic Doctrine behavior for `LAZY` collections. We should switch to `EXTRA_LAZY`: https://www.doctrine-project.org/projects/doctrine-orm/en/3.6/tutorials/extra-lazy-associations.html

Doing so will cause Doctrine to use a native SQL query for `Collection#count()` instead of initializing the collection and counting in memory. There should be no breaking side effects to doing so.

turegjorup (commented):
## Claude Code Review: Bugs and Performance Regressions

### BUG 1: ReleaseLoader — `activePromise` never cleared on cache hit

`assets/shared/release-loader.js:18-48`

```javascript
activePromise = new Promise((resolve, reject) => {
  if (latestFetchTimestamp + configFetchInterval >= nowTimestamp) {
    resolve(releaseData);       // ← resolves, but...
  } else {
    fetch(...)
      .finally(() => {
        activePromise = null;   // ← only clears here!
      });
  }
});
```

When the cache is hit (the `if` branch), `activePromise` is resolved but **never set back to `null`**. The `.finally()` only runs in the fetch path. This means:

1. First call → fetches → `.finally()` clears `activePromise`
2. Second call (within 5 min) → cache hit → resolves correctly → `activePromise` stays set
3. Third call (even after 5 min) → `if (activePromise)` is true → returns stale promise → **never checks the timestamp again**

The 5-minute refresh interval stops working after the first cache hit. The release data is effectively cached forever.

### BUG 2: ReleaseLoader — error path never resolves the promise

`assets/shared/release-loader.js:31-42`

```javascript
.catch((err) => {
  if (releaseData !== null) {
    resolve(releaseData);
  } else {
    console.warn("Could not find release.json. Returning defaults.");
    return {                    // ← return goes nowhere!
      releaseTimestamp: null,
      releaseVersion: null,
    };
  }
})
```

When the first fetch fails and `releaseData` is still `null`, the code `return`s a defaults object from the `.catch()` handler. But this return value is the return of the `.catch` callback — it doesn't call `resolve()`. The outer `new Promise` **never settles**. The caller hangs forever.

Should be `resolve({ releaseTimestamp: null, releaseVersion: null })`.

### PERFORMANCE: N+1 queries on collection counts (all three providers)

**`ScreenProvider.php:64`** — `$object->getScreenGroups()->count()`
**`ScreenGroupProvider.php:39`** — `$object->getScreens()->count()`
**`ScreenGroupProvider.php:40`** — `$object->getScreenGroupCampaigns()->count()`
**`PlaylistProvider.php:62`** — `$object->getPlaylistSlides()->count()`

All these ManyToMany/OneToMany relations use default `LAZY` fetch. Calling `count()` on an uninitialized lazy collection triggers `initialize()` which **loads all related entities into memory**, then counts them in PHP.

On a `GetCollection` endpoint with 10 items per page:
- **ScreenProvider**: 10 × `getScreenGroups()->count()` = 10 extra queries loading full ScreenGroup entities
- **ScreenGroupProvider**: 10 × `getScreens()->count()` + 10 × `getScreenGroupCampaigns()->count()` = 20 extra queries
- **PlaylistProvider**: 10 × `getPlaylistSlides()->count()` = 10 extra queries

Minimum fix: set `fetch: 'EXTRA_LAZY'` on these mappings so `count()` issues `SELECT COUNT(*)` instead of hydrating all entities. Better fix: use repository queries (like `getCampaignCountForScreen` does).

Source: `vendor/doctrine/orm/src/PersistentCollection.php:448` — the `count()` method only uses the collection persister (SQL COUNT) when `fetch === FETCH_EXTRA_LAZY`. Otherwise it falls through to `AbstractLazyCollection::count()` which calls `$this->initialize()`, loading all entities.

### PERFORMANCE: 2N queries for campaign counts per screen

**`ScreenProvider.php:55-56`**:
```php
$output->campaignsLength = $this->screenRepository->getCampaignCountForScreen($id);
$output->activeCampaignsLength = $this->screenRepository->getCampaignCountForScreen($id, true);
```

Each screen in a list executes **2 DQL queries** (total campaigns + active campaigns). For a 10-item page, that's 20 extra SQL queries. Could be combined into a single query returning both counts via `SUM(CASE ...)`, or batched for the whole page.

### MINOR: Modal buttons only show first page of results

`ScreensButton` in `groups-columns.jsx`, `SlidesButton` in `playlists-columns.jsx`, and `ScreenGroupsButton` in `screen-groups-button.jsx` all fetch only the first page from the API, but the button label shows the total count (e.g. `group.screensLength`). If a group has more items than one page, the modal shows a truncated list with no indication.

The `CampaignsButton` handles this correctly with recursive pagination via `getAllScreenGroups`/`getAllScreenCampaigns` helpers.

### MINOR: Dead parameters in `getGroupColumns`

`groups-columns.jsx:66`:
```javascript
function getGroupColumns({ apiCall, infoModalRedirect, infoModalTitle }) {
```

These destructured params are no longer used since `ScreensButton` handles everything internally.

---

**Summary**: Two bugs in `ReleaseLoader` (stale cache + hanging promise), significant N+1 query overhead on all three list providers, and pagination gap in three of the four modal buttons.

turegjorup (commented):
## Claude Code Test Review: Clarity and Coverage

### `testSlidesLength` (PlaylistsTest.php:392-440)

**Clarity**: Clean. The create → verify 0 → add → verify 1 → remove → verify 0 → cleanup flow reads well.

**Coverage gaps**:
- Only tests count of 0 and 1. Doesn't verify count > 1 (e.g. add 2 slides, verify `slidesLength` = 2).
- Only tests single-item `GET`. The optimization targets list endpoints (`GetCollection`). No assertion that `slidesLength` appears in `/v2/playlists?itemsPerPage=...` responses.

---

### `testScreensLength` (ScreenGroupsTest.php:138-178)

**Clarity**: Clean, same pattern.

**Coverage gaps**:
- **Doesn't test `campaignsLength` at all.** The PR adds both `screensLength` and `campaignsLength` to ScreenGroup DTO/Provider, but only `screensLength` is tested. `campaignsLength` on ScreenGroup has zero test coverage.
- Same as above: only tests count 0→1, no count >1, no collection endpoint.

**Suggestion**: Either expand this test or add a `testCampaignsLength` that creates a screen group, links a campaign, and verifies `campaignsLength` increments. Minimally, assert `campaignsLength` = 0 alongside `screensLength` = 0 in the initial state check.

---

### `testCampaignsLengthAndActiveCampaignsLength` (ScreensTest.php:261-397)

**Clarity**: Well-commented, logical progression. The comment on line 322-323 acknowledging the confusing endpoint semantics is helpful.

**Coverage gaps — most critical**:

1. **Campaigns via screen groups are not tested.** The `getCampaignCountForScreen` query aggregates campaigns from both:
   - Direct: `ScreenCampaign` (screen → campaign)
   - Indirect: `ScreenGroupCampaign` (screen → group → campaign)
   
   The test creates a screen group but adds **zero campaigns** to it. So `campaignsLength` stays at 2 (all direct). The indirect path — the harder part of the query — is completely untested.
   
   **Suggested addition**: After adding the screen to the screen group, link a campaign to that screen group and verify `campaignsLength` increments to 3.

2. **Deduplication not tested.** The query uses `COUNT(DISTINCT p.id)`. If the same campaign is linked both directly to the screen AND via a screen group, it should be counted once. This edge case is untested. A failing implementation that omits `DISTINCT` would still pass all current tests.

   **Suggested addition**: Link the active campaign (already direct) to the screen group as well. Verify `campaignsLength` stays the same (not double-counted).

3. **Length**: At 137 lines this test covers three independent concerns (direct campaigns, active/inactive distinction, screen groups count). Consider splitting into focused test methods.

---

### Cross-cutting suggestions

| Improvement | Applies to | Impact |
|---|---|---|
| Test on collection endpoint (`GetCollection`) | All three | Verifies the primary use case — list views are why these fields were added |
| Test count > 1 | `testScreensLength`, `testSlidesLength` | Catches off-by-one errors |
| Add `campaignsLength` coverage for ScreenGroup | `testScreensLength` | Currently zero coverage for this new field |
| Test indirect campaigns via screen group | `testCampaignsLengthAndActiveCampaignsLength` | The hardest part of the DQL query is untested |
| Test deduplication (DISTINCT) | `testCampaignsLengthAndActiveCampaignsLength` | A missing DISTINCT would pass all current tests |

The most impactful single improvement would be adding a test that links a campaign to a screen group, adds the screen to that group, and verifies the indirect campaign is counted — then links the same campaign directly too and verifies it's not double-counted. This covers both the indirect path and deduplication in one scenario.

turegjorup:
## Bug: `CampaignsButton` — inner promise rejections leave button in permanent loading state

**File:** `assets/admin/components/screen/util/campaigns-button.jsx:66-105`

### Problem

The `onClick` handler chains nested `.then()` calls but never **returns** the inner promises from outer `.then()` callbacks. The single `.catch(() => setLoading(false))` on line 105 only catches rejection from the outermost `getAllPages` call.

If any inner call rejects (`getAllScreenGroupCampaigns`, second `getAllPages`, or `getAllCampaigns`) — e.g. due to a network failure — `setLoading(false)` is never called and the button stays in a spinner state permanently.

```javascript
// Current (buggy) — lines 69-104:
.then((screenGroups) => {
    // ⚠️ Not returned — rejection won't reach outer .catch()
    getAllScreenGroupCampaigns(dispatch, screenGroupIds).then(
      (screenGroupCampaigns) => {
        getAllPages(...).then((screenCampaigns) => {
          // ...
          getAllCampaigns(...).then((allCampaigns) => {
            setCampaigns(allCampaigns);
            setLoading(false);
          });
        });
      },
    );
  })
.catch(() => setLoading(false));
```

### Fix

Add `return` before each inner promise so rejections propagate to the outer `.catch()`:

```javascript
.then((screenGroups) => {
    // ...
    return getAllScreenGroupCampaigns(dispatch, screenGroupIds).then(
      (screenGroupCampaigns) => {
        return getAllPages(...).then((screenCampaigns) => {
          // ...
          return getAllCampaigns(...).then((allCampaigns) => {
            setCampaigns(allCampaigns);
            setLoading(false);
          });
        });
      },
    );
  })
.catch(() => setLoading(false));
```

### Verified

A test (`assets/tests/admin/campaigns-button-promise-chain.spec.js`) was written that replicates the exact promise chain with controllable mocks. Results:

| Rejection point | Buggy code | Fixed code |
|---|---|---|
| `getAllPagesScreenGroups` (outermost) | ✅ loading clears | ✅ loading clears |
| `getAllScreenGroupCampaigns` (inner) | ❌ loading stuck | ✅ loading clears |
| `getAllPagesScreenCampaigns` (inner) | ❌ loading stuck | ✅ loading clears |
| `getAllCampaigns` (inner) | ❌ loading stuck | ✅ loading clears |

Test file is not committed — remove or commit at your discretion.

turegjorup (approved):

PR #387, opened 2026-04-17
Added vitest for frontend unit tests

tuj:
#### Link to issue

https://github.com/os2display/display-api-service/issues/380

#### Link to ticket

https://leantime.itkdev.dk/#/tickets/showTicket/7203

#### Description

Added vitest for frontend unit tests.
Rewrite playwright unit tests as vitest unit tests.

#### Checklist

### DIFF
--- a/.github/workflows/vitest.yaml
+++ b/.github/workflows/vitest.yaml
@@ -1,0 +1,23 @@
+on: pull_request
+
+name: Unit tests
+
+env:
+  COMPOSE_USER: runner
+
+jobs:
+  vitest:
+    name: Vitest
+    runs-on: ubuntu-latest
+    steps:
+      - name: Checkout
+        uses: actions/checkout@v4
+
+      - name: Setup network
+        run: docker network create frontend
+
+      - name: Install dependencies
+        run: docker compose run --rm node npm install
+
+      - name: Run unit tests
+        run: docker compose run --rm node npm run test:unit
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -34,6 +34,7 @@ All notable changes to this project will be documented in this file.
 - Optimized release data fetching.
 - Optimized list loading.
 - Removed fixture length check from test.
+- Added vitest for frontend unit tests.
 
 ### NB! Prior to 3.x the project was split into separate repositories
 
--- a/README.md
+++ b/README.md
@@ -306,15 +306,23 @@ task test:api
 
 ### Unit tests - Vitest
 
+Use Vitest for unit and component tests (pure functions, utilities, components with jsdom).
+
+Test files are located in `assets/tests/` alongside the Playwright tests.
+
+Vitest picks up `*.test.js` files.
+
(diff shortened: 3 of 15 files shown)


turegjorup:
## Code Review

### Overview
Introduces **Vitest** for fast JS unit testing alongside the existing Playwright suite. Renames 5 existing pure-function tests from `.spec.js` to `.test.js`, adds a new `id-from-path.test.js`, wires up a CI workflow, Taskfile target, and separate Playwright/Vitest test matchers so the two runners don't overlap. Verified locally: **all 39 tests pass in ~2s**.

### What's good
- Clean separation via `testMatch` in `playwright.config.ts` and `include` in `vitest.config.js` — no risk of double-running.
- Good regression coverage, particularly `campaigns-button-promise-chain.test.js` (promise chain `setLoading(false)` regression) and `id-from-path.test.js` (null/undefined/number edge cases).
- CI workflow (`vitest.yaml`) follows the existing Docker-based pattern.
- README + CHANGELOG + Taskfile all updated — low friction for contributors.

### Issues & suggestions

**1. Unused dependency: `@testing-library/react`** (`package.json`)
Added to devDependencies but not imported anywhere. `setup.js` only uses `@testing-library/jest-dom/vitest`. Either drop it (and probably `jsdom`) until a component test lands, or add at least one component test to justify the dep weight.

**2. `jsdom` env costs ~8s of setup for tests that don't need it** (`vitest.config.js`)
The test output shows `environment 8.06s` — all current tests are pure function tests on plain JS modules and would run in `node` env. Consider defaulting to `environment: "node"` and opt-in to jsdom per-file via `// @vitest-environment jsdom` once component tests arrive.

**3. Inconsistent test API style**
Five files use `test.describe` / `test` (Playwright-style, preserved from the rename — works in vitest as a compat alias but not officially documented); the new `id-from-path.test.js` uses idiomatic `describe` / `it`. Pick one and stick to it — I'd recommend `describe`/`it` for forward-compat safety.

**4. Timing-based flakiness in promise chain tests** (`campaigns-button-promise-chain.test.js`)
`await new Promise((resolve) => setTimeout(resolve, 50))` to wait for microtasks is brittle on slow CI. Better: have `onClick` return its promise chain so the test can `await` it directly, or use `await new Promise(setImmediate)` / `vi.waitFor`.

**5. README guidance on when to use Vitest vs Playwright** (`README.md`)
"…located in `assets/tests/` alongside the Playwright tests." A one-liner on *when to use which* (pure utilities → Vitest, UI/flow → Playwright) would help future contributors choose correctly.

### Risks
- Low. Changes are additive; existing Playwright tests are untouched beyond the matcher filter.
- The new devDep footprint is sizable (jsdom pulls ~20+ transitive deps) — worth trimming if not used.

### Verdict
Solid foundation, ready to merge after (optionally) trimming the unused `@testing-library/react` dep and unifying the `describe`/`test` style. Items 2, 4, 5 are follow-ups, not blockers.

turegjorup (approved):
Claude Code flagged a few things, but no blockers.

PR #419, opened 2026-05-01
CI cleanup: cache vendor + node_modules, --no-deps, doc each workflow, rework apispec

turegjorup:
#### Description

Stacks on top of #323. Independent cleanups, each in its own commit. No `docker-compose.yml` changes.

1. **Cache `vendor/` across PHP workflows** (`actions/cache@v5` keyed on `composer.lock`). Wall-time saving is small (Symfony auto-scripts run on every install regardless), but the cache avoids hitting Packagist on every CI run — kept as good citizenship.
2. **Skip mariadb on database-less workflows** — `--no-deps` on `docker compose run` for code-analysis, rector, twig-cs, php-cs-fixer, composer normalize/validate, and the prod composer-install validator. Compose file unchanged; phpunit/doctrine/playwright keep `depends_on`.
3. **Cache `node_modules/` in JS workflows** — `actions/cache@v5` keyed on `package-lock.json` for `vitest.yaml` and `playwright.yaml`.
4. **Drop no-op `strategy.fail-fast`** from seven non-matrix workflows.
5. **Drop the itk-dev tooling header, document every workflow uniformly** — eight workflows carried a stale "do not edit" header pointing at `itk-dev/devops_itkdev-docker`. The remaining eleven workflows had no header documentation; they get the same `### ### Title / Description / Assumptions` shape.
6. **Rework `apispec.yaml`** — adapts itksites' two-job api-spec workflow with one behaviour change: stay quiet when the spec is unchanged. Posts a single edited-in-place comment on (a) drift, (b) non-breaking changes (changelog), or (c) breaking changes (changelog + failure). Pinned `oasdiff-action` to `v0.0.44`, added a `paths:` filter, breaking-changes job depends on the export job. Excluded from the benchmark below — its scope changed.

#### Measured timings

Two CI runs back-to-back: the first populated caches; the second is the cache-hot baseline this PR's caching gives going forward.

| Workflow | Baseline (#323) | Cold (this PR) | Hot (this PR) | Hot vs Baseline |
|---|---|---|---|---|
| Test (phpunit) | 187s | 193s | 172s | **−15s** |
| Code Analysis | 57s | 52s | 45s | **−12s** |
| Rector | 56s | 53s | 44s | **−12s** |
| Composer (audit) | 51s | 43s | 40s | **−11s** |
| Composer install | 45s | 38s | 40s | −5s |
| Doctrine | 45s | 47s | 50s | within variance |
| Symfony PHP | 46s | 45s | 57s | within variance |
| Twig | 44s | 35s | 55s | within variance |
| Unit tests (vitest) | 34s | 48s | 39s | +5s (cache overhead) |

The wins are driven by `--no-deps` (mariadb skip) on database-less workflows, not by vendor caching itself — Symfony's auto-scripts (`cache:clear`, `assets:install`) run on every `composer install`, swamping any package-download saving. Vendor cache is kept anyway because every CI run otherwise pulls the same packages from Packagist.

`node_modules/` cache trims `npm install` from 32s to 19s on the install step, but ~3s cache save/restore overhead and runner variance net out a modest delta on the wall-time.

CI runner variance is ~10-15s for these workflows; single-run benchmarks at this scale are noisy.

#### Net

Per-PR savings on the slowest workflow (phpunit, the gating one): **~15s**. Less than the original estimate, but consistent across runs.

#### Notes

- `actions/cache@v5.0.5`, `oasdiff-action@v0.0.44`.
- Targets `feature/update-infrastructure-for-mono-repo` so the diff stays scoped.
- Possible follow-up (out of scope for this PR): `composer install --no-scripts` on static-analysis-only workflows would shave ~15s off each by skipping the auto-scripts. Needs a check that PHPStan etc. don't need the container dump.

#### Checklist

- [x] My code is covered by test cases. (CI itself is the test.)


.github/workflows/apispec.yaml
```diff
@@ -1,34 +1,160 @@
...
+              echo "The breaking changes action detected incompatible changes. Review the action logs for details."
+            fi
+          } > comment.md
+          gh pr comment ${{ github.event.pull_request.number }} \
+            --body-file comment.md \
+            --create-if-none --edit-last
 
-      - name: Check for changes in specifications (json)
-        run: git diff --diff-filter=ACMRT --exit-code public/api-spec-v2.json
+      - name: Fail if breaking changes detected
```
tuj, line 158: How do we handle if we want to merge breaking changes?

tuj (approved):
Approved with comment

.github/workflows/apispec.yaml
turegjorup: With comment on PR explaining motivation/reasoning.

PR #429, opened 2026-05-04
Fix Create Github Release workflow node_modules permissions

turegjorup:
## Summary

Fixes the `Create Github Release` workflow failing at `Cleanup before packaging` because root-owned files from the `node` compose service could not be removed by the runner.

- Add overridable `user` to the `node` service (defaults to `root` so local-dev is unchanged).
- Set `COMPOSE_NODE_USER` to the runner's UID:GID and pass `HOME=/tmp` so npm can write its cache when not running as root.
- Align `.dockerignore` with the workflow's `Cleanup before packaging` list (`/phpstan-baseline.neon`, `/playwright/`).

Why scoped to a new `COMPOSE_NODE_USER` instead of reusing `COMPOSE_USER`: the `node:24` image has no built-in non-root user matching the runner UID, while `itkdev/php8.4-fpm:alpine` does (via `runner`). Reusing `COMPOSE_USER` would require touching the eleven other workflows that already rely on it for phpfpm.

## Test plan

- [x] Verified on a fork: tag push of `0.0.0-test3` produced a successful workflow run (1m31s), `Cleanup before packaging` passed, the GitHub release was created and assets (`*.tar.gz`, `checksum.txt`) uploaded.
- [ ] Confirm local `docker compose run node …` still works unchanged (root default preserved).
- [ ] Maintainer to push a real semver tag against `release/3.0.0` after merge to verify end-to-end.

🤖 Generated with [Claude Code](https://claude.com/claude-code)

### DIFF
--- a/.dockerignore
+++ b/.dockerignore
@@ -26,6 +26,7 @@
 
 # CI / quality-tool configs (not consumed at runtime)
 /phpstan.dist.neon
+/phpstan-baseline.neon
 /phpunit.xml.dist
 /psalm.xml
 /psalm-baseline.xml
@@ -65,7 +66,7 @@ launch.json
 /test-results/
 /playwright-report/
 /blob-report/
-/playwright/.cache/
+/playwright/
 #< Playwright
 
 ###> vincentlanglet/twig-cs-fixer ###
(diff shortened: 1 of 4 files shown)


tuj (approved):
Markdown linting

PR #433, opened 2026-05-06
fix: Instant Book: Cleans up issue with caching

tuj:
#### Link to isse

https://github.com/os2display/display-api-service/issues/436

#### Link to ticket

https://leantime.itkdev.dk/#/tickets/showTicket/7208

#### Description

Cleans up issue with Instant Book caching and fixed other bugs.

#### Checklist

#### Changes

- Controller uses a typed DTO:
  - InteractiveController reads InteractiveSlideActionInput from request->attributes->get('data') instead of $request->toArray().
  - DTO gains implementationClass; data is now ?array.
  - Service method renamed parseRequestBody(array) → parseInteractiveSlideActionInput(InteractiveSlideActionInput).
  - API spec + RTK type regenerated.

- InstantBook bug fixes
  - intervalFree boundary fix: >/< → >=/<= so adjacent bookings don't count as overlap.
  - getBusyIntervals end-time was using start-time's timezone — now uses eventEnd['timeZone'].
  - Spam-protect key changed from slide->getId() to $resource (per-resource throttling).
  - Spam-protect logic was always-true (now+ttl > now); rewritten with a &$isFreshRequest flag set inside the cache miss callback.
  - durationMinutes validated against [15, 30, 60] instead of being passed straight to DateInterval.
  - Booking response: HTTP 409 → ConflictException, any >= 400 → NotAcceptableException (was silently returning the failed status).
  - CACHE_KEY_RESOURCES now actually expires (PT5M); was being cached forever.
  - Property typo keyValueService → keyVaultService.
  - Exception-message string interpolation fix in getValueFromInterval.

- Performance / cache
  - New getBusyIntervals cache (PT15M) — eliminates per-poll Graph calls.
  - Pagination via @odata.nextLink in getBusyIntervals (do/while, merges per scheduleId).
  - Pre-flight conflict check removed from quickBook — relies on Graph 409 instead.
  - Sibling cache writes moved out of the outer cache->get() callback into a new setCacheValue() helper (avoids nested cache mutation).
  - New validateResourceAccess() deduplicates ~15 lines of config/permission/feed checks shared by quickBookOptions and quickBook.

- Risk notes
  - 15-minute busy-intervals cache means availability shown in quickBookOptions can be up to 15 min stale; the actual booking call still 409s, so users get a correct error but an outdated picture.


src/InteractiveSlide/InstantBook.php
```diff
@@ -406,48 +398,53 @@ private function getBusyIntervals(string $token, array $resources, \DateTime $st
...
+                    $result[$scheduleId][] = [
+                        'startTime' => $start,
+                        'endTime' => $end,
+                    ];
+                }
             }
-        }
+
+            $url = $data['@odata.nextLink'] ?? null;
+        } while (null !== $url);
```
turegjorup: Non-blocking: Graph's `@odata.nextLink` is followed with `GET`, not a `POST` with the original body. Unlikely to fire on 1-hour windows, but if it ever does the second iteration will probably 4xx. Worth either a comment noting the assumption, or switching to `GET` (no body) when `$url` came from nextLink.

turegjorup (approved):
LGTM. Bug fixes (interval boundary, timezone, spam-protect, duration validation, error mapping) and the DTO refactor look correct.

One non-blocking nit, pre-existing but in a file touched here: `InteractiveServiceTest::testPerformAction` lines 86–91 are unreachable — `performAction()` on line 84 throws and ends the test, so the second scenario never runs, and the expected message `'Action not allowed'` doesn't match what `InstantBook::performAction()` throws (`'Action not supported'`). Worth splitting into two test methods or dropping the dead branch in a follow-up.

src/InteractiveSlide/InstantBook.php
tuj: Removed after inspecting spec

PR #434, opened 2026-05-06
feat: Instant Book: Added strategy where busy intervals are retrieved from calendar feed

tuj:
NB! Builds on https://github.com/os2display/display-api-service/pull/433

#### Link to issue

https://github.com/os2display/display-api-service/issues/436

#### Link to ticket

https://leantime.itkdev.dk/#/tickets/showTicket/7208

#### Description

- Added strategy to "Instant Book" where busy intervals are retrieved from calendar feed.
- Fixed instant booking template error handling.
- Fixed bugs when fetching intervals.
- Fixed calendar test issue following change to 10 min default update interval.

#### Checklist

### DIFF
--- a/.env
+++ b/.env
@@ -165,6 +165,11 @@ CALENDAR_API_FEED_SOURCE_CACHE_EXPIRE_SECONDS=300
 EVENTDATABASE_API_V2_CACHE_EXPIRE_SECONDS=300
 ###< Event Database Api V2 Feed Type ###
 
+###> Instant Book ###
+# Source of busy intervals data for instant booking interactive slides (graph or feed).
+INSTANT_BOOK_BUSY_INTERVALS_SOURCE=graph
+###< Instant Book ###
+
 ###> Admin configuration ###
 # API key for Rejseplanen (Danish journey planner) integration.
 ADMIN_REJSEPLANEN_APIKEY=
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -6,6 +6,7 @@ All notable changes to this project will be documented in this file.
 
 - Added `INSTANT_BOOK_BUSY_INTERVALS_SOURCE` to select between Graph and the slide's calendar feed as the source of busy
   intervals for InstantBook.
+- Changed polling interval for instant booking template.
 - Refactored InteractiveController to use a typed `InteractiveSlideActionInput` DTO; regenerated API spec and RTK types.
 - Fixed multiple InstantBook bugs: interval boundary overlap, busy-interval timezone, per-resource spam-protect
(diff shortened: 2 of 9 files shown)


turegjorup:
## Local review follow-up

Worked through the review findings on this branch. Five commits pushed on top of the original head (`7b4318b3..3b9f1e64`).

### Fixed

| Concern | Resolution | Commit |
|---|---|---|
| 406 (Not Acceptable) for `INSTANT_BOOK_BUSY_INTERVALS_SOURCE` misconfig | Constructor guard throws `\InvalidArgumentException` → 500; `default` arm dropped as unreachable | `ca263e37` |
| 406 for slide-config errors (null feed / null source / non-calendar type) | `Symfony\…\UnprocessableEntityHttpException` → 422 (admin-fixable data state, not Accept-header negotiation) | `557fc4fb` |
| Pre-booking Graph call looked like wasted overhead | Comment rewritten: the resources here are *not* configured to block conflicts at the Graph API level — a conflicting POST would silently succeed — so the pre-check is **required for correctness**, not an optimisation. The 409 catch is a backstop for resources that do reject conflicts | `a7ac945e` |
| `CACHE_KEY_BUSY_INTERVALS_PREFIX` was a fixed key — different slides could read each other's data for up to 15 min | Extracted to `getBusyIntervalsCached()`; key now built from sorted resource set + from/to at minute precision | `91d028ed` |
| No test coverage for the new conflict pre-check | Extracted to `assertSlotFree()`; added tests for busy → `ConflictException` and empty → pass | `3b9f1e64` |

### Verified safe (no change needed)

- **Frontend `apiUrl` removal in `calendar-single-booking.jsx`** — verified the change is a bug fix, not a regression. `slide["@id"]` is an absolute path (`/v2/slides/<ulid>`) per API Platform's default `url_generation_strategy = 1` (`ABSOLUTE_PATH`); confirmed by existing test assertions in `LayoutsTest.php`. The previous `${apiUrl}${slide["@id"]}/action` produced broken URLs because `apiUrl` came from `localStorage` populated with `configData.apiEndpoint`, which `ClientConfigController` does not actually return — so `apiUrl` was the literal string `"undefined"` in any non-fallback config.

### Out of scope / FYI

- **`KobaFeedType` timestamp shape** — dropped; the feed type is `@deprecated`, defensive code not warranted.
- **Polling cadence 5s → 60s** — UX choice, documented in CHANGELOG.
- **`phpunit.xml.dist` `disabled` → `weak`** — minor scope creep; could be mentioned in the description or split if you want cleaner history.

Tooling: validated via `task coding-standards:php:check`, `task code-analysis`, `task test:api -- tests/Interactive/InstantBookTest.php` after each change. All clean.

turegjorup (approved):

PR #458, opened 2026-05-19
feat: Made upload sizes configurable

tuj:
#### Link to issue

https://github.com/os2display/display-api-service/issues/392

#### Link to ticket

https://leantime.itkdev.dk/#/tickets/showTicket/7259

#### Description

Made upload sizes configurable.

#### Checklist


turegjorup (approved):

turegjorup:
## Response to review findings

Updated to reflect commits landed on this branch.

### Addressed

| Finding | Commit |
|---|---|
| `MediaController.php:41` (now `MediaFileValidator.php`) — `'0Mi'` rejects every byte when env coerces to 0 | `6d0326c3` — `__construct` throws `\InvalidArgumentException` naming `MEDIA_MAX_UPLOAD_SIZE_MB` on any non-positive value; dataprovider test covers `0` and `-1` |
| `.env:111` vs README — `cache:clear` contradiction | `a016403e` — `.env` note now states the value is read per-request; also documents the MiB unit and nginx/php-fpm alignment requirement |
| `file-selector.jsx:39` — falsy-zero guard ignores server value `0` | `bd25feac` — replaced with `Number.isInteger(value) && value > 0 ? value : 200` |
| `file-selector.jsx:34` — dropzone renders permissive 200 MB default before config loads | `bd25feac` — initial state `null`; dropzone hidden behind `file-selector.loading` placeholder until config resolves; new vitest case covers the transition |
| `admin-config-loader.js:28` — fallback resolved but not cached, every consumer mount re-fired the failing fetch | `b5f594c9` — defaults hoisted to module-level `DEFAULT_CONFIG`; catch path now assigns `configData = DEFAULT_CONFIG` |
| `admin-config-loader.js:10` — latent stale `activePromise` on the cache-hit branch | `b5f594c9` — rewrite removes the wrapping `new Promise(...)`; `activePromise` only set during fetch, cleared in `.finally` |
| `services.yaml:72` — `AdminConfigController` arguments block shadowed the global bind for 6/7 params | `a8515445` — removed the global `$mediaMaxUploadSizeMb` bind entirely; explicit per-service wiring for `AdminConfigController` and the new `MediaFileValidator` |
| `MediaController.php:38` — validation order surfaced size/mime error before missing-title | `a8515445` — title/description/license are read before the entity validator runs, so `MediaException('Missing request parameter: title')` fires first |
| `MediaTest.php:166` — stale `Media::loadValidatorMetadata` comment | `a8515445` — updated to reference the new `#[MediaFile]` attribute on `Media::$file` |

### Structural change (`a8515445`)

The size/mime constraint moved off the controller and back onto the entity via a custom `App\Validator\MediaFile` constraint + validator (the API Platform [`MinimalProperties`](https://api-platform.com/docs/symfony/validation/) pattern). The controller now calls `$this->validator->validate($media)` against the assembled entity; the validator reads `MEDIA_MAX_UPLOAD_SIZE_MB` via DI. Restores the entity as the single source of truth for upload invariants and removes the inline `new Assert\File(...)` from the controller. Also dropped the redundant `env(MEDIA_MAX_UPLOAD_SIZE_MB): '200'` parameter fallback in `services.yaml` — `.env` is the sole default, consistent with how other env-driven params are handled.

### Reclassified after investigation

- **`MediaController.php:30` — no handling of `UPLOAD_ERR_INI_SIZE` / `UPLOAD_ERR_FORM_SIZE`.** Original finding was overstated. Symfony's `Assert\File` already branches on `UploadedFile::getError()` and emits a specific violation for every PHP upload error code (`vendor/symfony/validator/Constraints/FileValidator.php:54-116`); our `MediaFileValidator` delegates to it, so the `upload_max_filesize` case surfaces as a `422` with a clear "file is too large" message for free. Only `post_max_size` remains uncovered — PHP discards the body before Symfony's kernel runs, so there's no `UploadedFile` for any validator to inspect. With the PR's defaults (PHP-FPM ≥ nginx ≥ app at 200 MiB), nginx is the strictest layer and rejects with 413 before PHP runs, making this hard to hit in practice. Leaving the misleading "file is required" 400 for the narrow `post_max_size` case as a known gap.

### Downgraded

- **`docker-compose.yml:72` — `NGINX_MAX_BODY_SIZE: 5M`** — confirmed dev-only stack. Production runs the image built from `infrastructure/nginx/Dockerfile` whose `NGINX_MAX_BODY_SIZE=200m` default is already aligned. Local dev value is a dev-UX inconvenience for testing the new feature locally, not a release blocker.

### Not addressed (deferred / out of scope)

Listed for tracking. Happy to pull any of these into this PR if you'd like:

- **`MediaTest.php:169` — skip-gate reads `$_ENV` directly.** Silent skip if dotenv isn't populated. Should read via container/parameter instead.
- **`MediaController` — `200Mi` (binary) vs historical `200000k` (decimal) ≈ 4.86% looser default.** Worth either a CHANGELOG note or switching to `200M`.
- **`file-dropzone.jsx:29` — error message mixes decimal MB display (`/1000000`) with binary MiB threshold (`*1024*1024`).** Near boundary yields self-contradictory "X MB > X MB".
- **`infrastructure/nginx/Dockerfile:31` — nginx body size hard-coded `200m`, not derived from `MEDIA_MAX_UPLOAD_SIZE_MB`.** Raising the app limit silently misaligns prod.

PR #477, opened 2026-06-02
feat(logging): [3/4] PHPStan logging gate + silent-catch migration

turegjorup:
## PR 3 of 4 — Structured logging

Builds on **#479** (semconv/redaction/exception processor). **Base is `feature/structured-logging-semconv`** until the earlier PRs merge, then retargets to `release/3.0.0`.

Adds the CI gate that enforces the logging conventions **and** migrates every flagged site in the same PR, so the rules and the fixes land together. **Zero** new `phpstan-baseline.neon` entries for the three identifiers.

### Rules (`tools/phpstan/`)

- **`NoSilentCatchRule`** (`logging.silentCatch`) — flags a `catch` that doesn't rethrow, log via PSR-3, call `error_log`/`trigger_error`, **or report the failure to the console** (`SymfonyStyle`/`OutputInterface`). The console-output exemption lives in the rule (not a `src/Command/` path exclusion in `phpstan.dist.neon`) so a genuinely empty catch *inside* a command — which hides failures in unattended cron/Messenger runs — is still flagged.
- **`NoInterpolatedLogMessageRule`** (`logging.interpolatedLogMessage`) — the log message must be a static string; variable data belongs in the context array. `{placeholder}` braces resolved from context are allowed.
- **`ExceptionInContextMustUseExceptionKeyRule`** (`logging.exceptionContextKey`) — a `\Throwable` in the context array must use the `exception` key (so `ExceptionContextProcessor` serialises it).

> A correctness note for reviewers: the plan's draft `NoSilentCatchRule` matched `Stmt\Throw_`, but php-parser v5 (PHPStan 2.x) emits `Expr\Throw_` — so rethrowing catches were initially mis-flagged. Fixed to `Expr\Throw_`.

### Silent-catch migration (every flagged site resolved)

- **Logged** (channel-bound logger injected, exception under the `exception` key): `AuthOidcController`, `AuthScreenBindController`, `ColiboFeedType`, `InstantBook` (×2), `FeedService`, `FeedProvider`, `ThemeProvider`, `UserProvider`.
- **Interpolation fixed** (already logging): `NotifiedFeedType` (×2), `RssFeedType`.
- **Annotated** (genuinely intentional, with a one-line reason): the optional OIDC end-session URL, per-booking time parsing in `BrndFeedType`, the documented read-degradation in `FeedSourceProvider`, the log-enrichment guard in `RequestContextProcessor`, `RRuleValidator` (surfaces a constraint violation), and the dev-only `RRuleProvider` fixture.

Command catches that already report via `$io->error(...)` + a non-zero exit are handled by the rule's console exemption — no annotations needed.

### Verification

- `composer code-analysis` — green with the three new rules
- `composer coding-standards-check` — green
- `composer test` — 178 tests, 708 assertions, green (constructor changes propagated to `AuthOidcControllerTest`, `InstantBookTest`, `FeedServiceTest`)
- Scratch-file check: each identifier fires on a violating snippet; the `$io->error` console path does **not** fire — confirming the exemption.

🤖 Generated with [Claude Code](https://claude.com/claude-code)


composer.json
```diff
@@ -29,7 +29,7 @@
         "nelmio/cors-bundle": "^2.1",
         "pentatrion/vite-bundle": "^8.1",
         "phpdocumentor/reflection-docblock": "^5.2",
-        "phpstan/phpdoc-parser": "^1.24",
+        "phpstan/phpdoc-parser": "^2.0",
```
tuj: This should be a dev requirement

tuj (changes requested):

composer.json
turegjorup: Moved

tuj (approved):

